Issue #12, opened 2016-11-04
Multi-compiler support


gpoitch:
For webpack configs that return arrays, only one of the bundles has a report generated.  
Throws an EADDRINUSE in server mode, presumably trying to launch both.

EDIT: does work correctly in 'static' mode.

## Edit by @valscion

**Help contribute a fix for this issue here: https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/281**

mkarajohn:
I have a `webpack.config` that exports 4 other configs. I am using the plugin in each of them separately and I get 4 report pages. How are you using it?

gdub22:
Essentially just this (stripped down) webpack.config.js:
```js
module.exports = function () {
  const config1 = {
    // ...
    plugins: [new BundleAnalyzerPlugin()]
  }

  const config2 = {
    // ...
    plugins: [new BundleAnalyzerPlugin()]
  }

  return [config1, config2]
}
```
It tries to start 2 servers and crashes `Error: listen EADDRINUSE :::8888`.
Fixable by defining a different `analyzerPort` for each.  
Feel free to close, but might be fixable to work out of the box.

valscion:
One way to do this is exactly to set a different `analyzerPort` for each instance, or if using `static` mode, to set different `reportFilename` options for each instance.

The way `assets-webpack-plugin` works in multi-compiler mode is by only creating one instance of the plugin and adding that instance to the `plugins` array: https://github.com/kossnocorp/assets-webpack-plugin/tree/v3.5.1#using-in-multi-compiler-mode

That could be a way to support multi-compiler mode in the future, as long as the implementation of this plugin would support it. Right now it doesn't seem to be so, as the reporter cannot combine results from multiple `'done'` plguin events yet.

lependu:
Hi! 
Just for the record, the `next-bundle-analyzer` has hardcoded config which does not suit me.
But I stole its code and make it work in server mode:
```javascript
// nex.config.js
module.exports = {
  webpack: (config, options) => {
    if (ANALYZE) {
      const { isServer } = options
      config.plugins.push(new BundleAnalyzerPlugin({
        analyzerMode: 'server',
        //  The host changed from 127.0.0.0 to make it work when running it in docker container. Maybe you do not need that change.
        analyzerHost: '0.0.0.0', 
        analyzerPort: isServer ? '8881': '8882',
        openAnalyzer: false
      }))
    }

    return config
  }
```

[...]

PR #14, opened 2016-11-07
Assign a new valid port if port is in use

mkarajohn:
It was requested in #12 and I am also using a similar setup, as the one described there, so I thought I'd adjust `viewer.js` to retry to `.listen()` to a new port if the currently assigned one is already in use.

# Possible improvements:

Since now it retries by simply using the next port, if you have, say `k` instances of the plugin (all using port 8888 for example), for any nth instance it will invoke `app.listen()` `n` times, until it gets one port that's unused. Maybe we could be storing already used ports and assigning a new one by checking them first, so that we just do one `app.listen()` call, which will be guaranteed to succeed. Then again maybe that's an overkill anyway.

P.S. I am not sure which branch I was supposed to use as base.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -49,18 +49,30 @@ function startServer(bundleStats, opts) {
     });
   });
 
-  return app.listen(opts.port, () => {
-    const url = `http://localhost:${opts.port}`;
+  function listen(port) {
+    const server = app.listen(port, () => {
+      const url = `http://localhost:${port}`;
 
-    console.log(
-      `${bold('Webpack Bundle Analyzer')} is started at ${bold(url)}\n` +
-      `Use ${bold('Ctrl+C')} to close it`
-    );
+      console.log(
+          `${bold('Webpack Bundle Analyzer')} is started at ${bold(url)}\n` +
+          `Use ${bold('Ctrl+C')} to close it`
+      );
 
-    if (opts.openBrowser) {
-      opener(url);
-    }
-  });
+      if (opts.openBrowser) {
+        opener(url);
+      }
+    });
+
+    server.once('error', (err) => {
+      const nextPort = port + 1;
+      console.warn(`Unable to listen on port ${port}`);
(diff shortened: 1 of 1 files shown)


mkarajohn:
Hey @th0r, any thoughts on this, when you have time maybe?

th0r:
Sorry for a long delay - didn't have time to look at it.
The main reason for this issue (with ports) is because the plugin doesn't have support for multi-configs and there is an issue about it (#12).
This PR fixes a cause but I would like to fix a root of problem. What I'm thinking of is a possibility to include statistics from multiple configs into single analyzer report so it will still need one http port to start single server.
If I won't find a way to do it I'm happy to merge this PR.

th0r:
Also you can use `port: 0` that will tell Node.js to choose random port ([doc link](https://nodejs.org/dist/latest-v6.x/docs/api/http.html#http_server_listen_port_hostname_backlog_callback)).

valscion:
Hi, thanks for your contributions @mkarajohn! Sorry that we've dragged this PR for so long.

As @th0r said, you could use `analyzerPort: 0` to tell Node.js to choose a random port. It's not the prettiest solution out there, but it will have to do for now.

I'll close this PR as I don't think we will be going along with this specific kind of fix. Sorry

GodofHope:
I thought you shoulld accept this PR.
I am using Vue Cli 3 ,when using moderm build , the cli will build twice, once with a legacy build,  the other is a moderm build. 
And I think this tool does need an auto fix when the default port is not avaliable!!! 
You should consider the user has many 8000+ ports in use.

GodofHope:
I tried `{analyzerPort: 0}`  but it doesn't work. I wonder if there is any other solutions.

th0r:
> I used`{analyzerPort: 0}` as you mentioned. I suggest a hint added in the readme.md. Cause you should not think everybody is familar with node APIs...

I think we can create an alias `analyzerPort: 'auto'` that will convert the value to `0` under the hood.

PR #20, opened 2016-11-19
feat: support all imports

gdi2290:
```es6
import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer'
import * as BundleAnalyzerPlugin from 'webpack-bundle-analyzer'
import BundleAnalyzerPlugin from 'webpack-bundle-analyzer'

const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,6 @@
 const { start } = require('./viewer');
+const BundleAnalyzerPlugin = require('./BundleAnalyzerPlugin');
 
-module.exports = {
-  start,
-  BundleAnalyzerPlugin: require('./BundleAnalyzerPlugin')
-};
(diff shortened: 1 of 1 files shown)


valscion:
It might be better to only have one way of importing the plugin for future backwards compatibility.

It would be impossible to add some new features that could be directly imported if this were to be merged. Making sure `webpack-bundle-analyzer` would be backwards-compatible with this change sounds like an unnecessarily tough task.

th0r:
Agree with @valscion here - it'll just confuse people.

PR #21, opened 2016-11-24
setTimeout -> setImmediate

somebody32:
If you run webpack not directly but through a wrapper (like grunt-webpack) then using `setTimeout` generates race conditions: sometimes grunt is killing webpack thinking that everything went good and bundle-analyze is not firing at all.

Changing that to `setImmediate` fixes the problem and still placing all the logs to the bottom


th0r:
Thanks!

th0r:
In v1.5.4

PR #31, opened 2016-12-01
Fix case when bundleDir is '/'

carlosbaraza:
For some reason in my repository, bundleDir is set to '/', causing the bundle analyzer plugin to try to write in '/report.html', therefore breaking because it was no permissions.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -87,7 +87,8 @@ function generateReport(bundleStats, opts) {
       let reportFilepath = reportFilename;
 
       if (!path.isAbsolute(reportFilepath)) {
-        reportFilepath = path.resolve(bundleDir || process.cwd(), reportFilepath);
+        const outputPath = bundleDir !== '/' && bundleDir || process.cwd();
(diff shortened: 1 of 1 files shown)


th0r:
@carlosbaraza Well, seems like you're fixing the cause and not the root here.
Let's better find out why `bundleDir` equals to `/`.
How do you use it? As plugin?

carlosbaraza:
Haha sorry th0r, I agree it was a premature PR. My bundleDir is / due to
the parallel webpack plugin which seems to do that.

I don't have a proper test to reproduce the issue. I will have to spend
some time tomorrow trying to reproduce it. Hopefully I will have some time.

Otherwise I found a solution setting the filename as an absolute path:

```javascript
reportFilename: __dirname + '/report.html'
```

On 1 Dec 2016 19:25, "Yuriy Grunin" <notifications@github.com> wrote:
>
> @carlosbaraza Well, seems like you're fixing the cause and not the root
here.

th0r:
@carlosbaraza If it still bothers for you open a separate issue please.

Issue #32, opened 2016-12-07
Split analyzer from the treemap UI code


valscion:
This issue is a place to discuss and track a possible separation of webpack analyzer plugin code from the reporter UI code.

This idea was triggered by a proposal to add [sunburst charts to the UI](https://github.com/th0r/webpack-bundle-analyzer/pull/25), where @th0r and me eventually thought it might be interesting to split these two concerns to distinct packages:

> > Or that the UI that is opened could be configurable, so that I could build a totally separate npm package that would only contain the UI parts that could be plugged into this packages webpack parts.
>
> That's a very interesting idea and, at first glance, it won't be too hard to split analyzer from report UI.
> 
> We can move all current UI code from `webpack-bundle-analyzer` into, say, `webpack-bundle-analyzer-treemap-reporter`.
> `webpack-bundle-analyzer` will be responsible only for analyzing bundle and feeding this info to reporter-plugins whose responsibility will be showing it in any way.
> 
> And the nice thing is reporter is not limited to some UI stuff - it can do anything it want with provided data!

valscion:
I didn't hack on this on Christmas break but instead hacked on something completely different from work related matters to get a good vacation 🎅 . If I'll get to this, I'll let you know.

valscion:
I'm gonna give this a go now. @bitpshr, would it be OK to try to get [your beautiful sunburst chart](https://github.com/th0r/webpack-bundle-analyzer/compare/master...bitpshr:sunburst-chart?expand=1) to work first? It could be created as a `webpack-bundle-analyzer-reporter-sunburst` for example, if we'd want to try to prefix all the different reporters with `webpack-bundle-analyzer-reporter-` to ease discoverability.

valscion:
Continuing from the refactoring done in #40, the [current situation](https://github.com/th0r/webpack-bundle-analyzer/tree/84bb739330f5e7b2b6306f7ff2613adff07b8ad0) has a folder structure looking like this:

```
webpack-bundle-analyzer/
├── client
│   ├── components
│   └── vendor
├── lib
│   ├── bin
│   ├── chartData
│   └── viewer
├── node_modules
├── public
├── src
│   ├── bin
│   ├── chartData
│   └── viewer
├── test
│   ├── bundles
│   ├── src
│   └── stats
└── views
```

The only things reaching into top-level `client`, `public` and `views` folders are the root-level `webpack.config.js` and code under `src/viewer`.

[...]

PR #40, opened 2017-01-21
Refactor code to allow splitting bundle data extraction from UI code

valscion:
The goal of this PR is to split the UI code into a separate package that can be leveraged by the current `webpack-bundle-analyzer` code.

See #32 for more reasoning


valscion:
So far I have moved files around `src/` directory so that I could get a better focus on what files require each other. I hope that these file locations were internal to this module and weren't part of the public API.

Each commit up to this point has passed tests, which can be seen from https://travis-ci.org/valscion/webpack-bundle-analyzer/builds, too

I was wondering, why do some files require other modules via the `lib/` build directory? I understand why the tests might want to do that, but not why the source files do that.

I'll change all the non-test requires to point to files inside `src` directory. It should be easy-ish to revert, if there is a good reason to still require from `lib`.

valscion:
I've first had to refactor the current code to get it to a state where I feel more confident in splitting out distinct parts. I try to keep the steps small so that following along one commit a time should make for a pleasant review experience.

Please let me know if you have trouble understanding anything I've done. I'm so far targeting to create breathing room for the separation, and I'm not yet 100% sure what the code will look like once I'm done.

valscion:
At this point in the refactoring, I'm starting to miss the lovely [`flow` types](https://github.com/facebook/flow) as all the various option parameter defaults and plumbing is starting to get confusing again.

What is your stance on static typing, @th0r? Would you be fine if I'd introduce some gradual typing to the `webpack-bundle-analyzer` codebase so that touching some function parameters wouldn't so easily break accidentally?

valscion:
Huh? Why did the 640098b commit build pass? https://travis-ci.org/th0r/webpack-bundle-analyzer/builds/194015131

It should've been broken! It referenced a file that got moved to a different folder:

```js
// test/viewer/parseUtils.js
const { parseBundle } = require('../../lib/viewer/parseUtils');
```

😕  It should've been this:

```js
const { parseBundle } = require('../../lib/chartData/parseUtils');
```

Really strange.

valscion:
What's the reason behind not requiring `bundleDir` to be always present, but having [this check in `src/viewer.js`](https://github.com/th0r/webpack-bundle-analyzer/blob/a6dacbdd1b03c50feffadc14134c346c0f6d7178/src/viewer.js#L90)?

```js
      if (!path.isAbsolute(reportFilepath)) {
        reportFilepath = path.resolve(bundleDir || process.cwd(), reportFilepath);
      }
```

I see you also check for its presence in [`src/analyzer.js` module's `getViewerData` function](https://github.com/th0r/webpack-bundle-analyzer/blob/a6dacbdd1b03c50feffadc14134c346c0f6d7178/src/analyzer.js#L37-L42):

```js
  // Trying to parse bundle assets and get real module sizes if `bundleDir` is provided
  let parsedModuleSizes = null;
  let bundlesSources = {};
  let parsedModules = {};

  if (bundleDir) {
```

The `process.cwd()` defaulting seems to have been first introduced in a [commit adding "static report" mode](https://github.com/th0r/webpack-bundle-analyzer/commit/d64445059d2ffaf590d1b9500563c1c8f931abc2) and the protective `if(bundleDir)` in `getViewerData` has been there ever since [the initial commit](2916cb62f57d69708add60edf96286bf7d79c28a).

The `bundleDir` value can come in via only two ways, that are more easily spotted in this PR:

1. From the **plugin instance** [ getting it from `this.compiler.outputPath`](https://github.com/th0r/webpack-bundle-analyzer/blob/fec8bb8677590c053153dacee22879b24b008283/src/BundleAnalyzerPlugin.js#L84)
2. From the **CLI** [arguments](https://github.com/th0r/webpack-bundle-analyzer/blob/fec8bb8677590c053153dacee22879b24b008283/src/bin/analyzer.js#L57) or [inferring it from the bundle stat file location](https://github.com/th0r/webpack-bundle-analyzer/blob/fec8bb8677590c053153dacee22879b24b008283/src/bin/analyzer.js#L66)

It seems like there shouldn't be any scenario where the `bundleDir` should be `null` or `undefined` after these steps. If I haven't misunderstood, the `this.compiler.outputPath` comes from `output.path` webpack config value which in turn [is required by webpack to work](https://webpack.js.org/concepts/output/#usage).

So it should be fairly safe to remove the overly protective check from the code. It would make my life a bit easier as I'd have one less thing to worry about :smile:

valscion:
I went and removed the `bundleDir` optionality in the last commit. As the indentation changed, it might be best viewed with whitespace turned off using the special `?w=1` query parameter in the URL: https://github.com/th0r/webpack-bundle-analyzer/pull/40/commits/21385bf887d241f110bc73c877d6dcd801345226?w=1

valscion:
I'll tackle the actual splitting of UI and chart data extraction in a new PR so that we can focus on only the refactoring here. Otherwise this PR risks becoming too big and the review would get waaaaay too cumbersome.

Let me know if you have _any_ questions! I'm very open in discussing any approaches I've taken. If you've got critique, I want to hear it! :smile:

valscion:
Hiya @th0r, have you had any time to look into this? If you're having trouble maintaining this project, I could help you. It'd be a shame for this project to wither as it is so useful to many :)

valscion:
Hi @th0r, would you be able to review this? It'd be nice to get this refactor merged so that there would be less merge conflicts in the future.

I'm fairly confident that this does not introduce any changes or backwards-incompatibilities. `webpack-bundle-analyzer` has never seemed to support reaching to the insides of this plugin, so moving things around should also be safe.

I would like to continue figuring out a good approach on splitting the view code altogether, but that I would do in another place :)

The current file hierarchy looks like this now:

```
webpack-bundle-analyzer/
├── client
│   ├── components
│   └── vendor
├── lib
│   ├── bin
│   ├── chartData
│   └── viewer
├── node_modules
├── public
├── src
│   ├── bin
│   ├── chartData
│   └── viewer
├── test
│   ├── bundles
│   ├── src
│   └── stats
└── views
```

The only things reaching into top-level `client`, `public` and `views` folders are the root-level `webpack.config.js` and code under `src/viewer`.

**Those folders and the webpack config form the implementation of the current treemap reporter.**

With this in place I have a much better situation with figuring out how to split the reporter away from the data generation.

valscion:
I'll close this in favor of #81 that features a much smaller amount of code changes.

Issue #29, opened 2016-11-28
Could't analyze webpack bundle


drcmda:
#### Issue description

I am getting this message in the end of the webpack emitts:

    TypeError: Cannot read property 'raw' of undefined Could't find any javascript bundles in provided stats file

#### Technical info
* Webpack Bundle Analyzer version: 2.1.1
* Webpack version: 2.1.0-beta.27
* Node.js version:  v7.2.0
* npm/yarn version: 3.10.5/0.17.9
* OS: Windows 10

#### Debug info
How do you use this module?

Plugin

If plugin, what options were provided? 

None

What other Webpack plugins were used?

    new webpack.LoaderOptionsPlugin({
        options: { worker: { output: { filename: "build/worker.js" } } } }),
    new webpack.HotModuleReplacementPlugin(),
    new webpack.NamedModulesPlugin(),
    new BundleAnalyzerPlugin()

[stats.zip](https://github.com/th0r/webpack-bundle-analyzer/files/616864/stats.zip)

th0r:
> Technical info

What are the actual versions behind the `latest`?

drcmda:
Added it :)

th0r:
Is `2.1.1` the only version of analyzer you tried or it worked before?

drcmda:
2.1.0 the same. I haven't known about it before.

kMeillet:
Same problem with :

```js
{
    "webpack": "^2.2.0-rc.0",
    "webpack-bundle-analyzer": "^2.1.1", 
}
```

ganapativs:
Same issue (Could't find any javascript bundles in provided stats file).

"webpack": "2.1.0-beta.20",
"webpack-bundle-analyzer": "^2.2.0"

christianalfoni:
Yeah, same with Webpack 2.1.0-beta.28... maybe a change to stats file? Looking into it now

christianalfoni:
Seems to be fixed here: https://github.com/gbakernet/webpack-bundle-analyzer/commit/b033c3ac7888d651686ac733fe7af92e54a18a5d

It worked on my end, though I have an issue with parsing bundles... so can not get gzip etc. stuff

alifr:
I'm not getting the TypeError, but I have a similar issue concerning the "Couldn't parse bundle asset..." and not getting the gzip stats. All my modules are AMD as we recently switched from requireJS.
"webpack": "^2.2.0-rc.0"
[...]

PR #42, opened 2017-02-01
Allow reports to be analyzed even with incomplete data

valscion:
This partially addresses #29 by allowing the report to be generated at all in the first place. Now only the parts of the bundle that failed to be analyzed don't contain sizes for Gzip or parsed parts, but will still render successfully.

I suspect this will also help in debugging the issue further a lot.

partial-info-but-still-works

### DIFF
--- a/src/analyzer.js
+++ b/src/analyzer.js
@@ -86,7 +86,7 @@ function getViewerData(bundleStats, bundleDir, opts) {
     asset.modules = _(bundleStats.modules)
       .filter(statModule => assetHasModule(statAsset, statModule))
       .each(statModule => {
-        if (parsedModuleSizes) {
+        if (parsedModuleSizes && parsedModuleSizes[statModule.id]) {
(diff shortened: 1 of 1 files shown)


valscion:
This solution allows the build to succeed in those cases when the modules couldn't be extracted correctly from the compiled bundles.

It might be useful to warn about this failure mode, though. The modules that don't have gzip nor parsed sizes are not sized appropriately in the treemap as they have nothing to compare against so every module takes the same space in all but "stat" based sorting.

#44 should fix the root cause properly. The webpack chunk module template doesn't seem to have any other new, unaddressed cases than what #44 adds support for: 

* https://github.com/webpack/webpack/blob/v1.14.0/lib/Template.js#L79
* https://github.com/webpack/webpack/blob/v2.2.1/lib/Template.js#L94

th0r:
I think #44 fixes the root of the issue so there is no need in this PR any more.

PR #44, opened 2017-02-04
Support extra bundles with modules inside Array(N).concat([])

valscion:
Normally, a chunk starts with something like this:

```js
webpackJsonp_adhaghasd([0,2],[ /* modules here */ ])
```

..or like this:

```js
!function(e){/* webpack bootstrap */}([ /* modules here */])
```

But at least in @matteoantoci's case in https://github.com/th0r/webpack-bundle-analyzer/issues/29#issuecomment-276680137, the troublesome chunk starts like this:

```js
webpackJsonplmn_ui_obe__name_Init([0],Array(104).concat([ /* modules here */]))
```

This case was not handled by the utility parsing the module object back from the compiled source using AST parser, [`parseUtils.js`](https://github.com/th0r/webpack-bundle-analyzer/blob/v2.2.0/src/parseUtils.js)

That module template comes from this line in the webpack source code: https://github.com/webpack/webpack/blob/v1.14.0/lib/Template.js#L91

It seems like this is some sort of an optimization inside webpack that allows the modules to be listed in chunks without having to specify hundreds of "empty" elements before.

This PR should hopefully fix #29 with a proper solution (unlike the quickfix done in #42).

**I'm asking all that had the `TypeError` over at #29 to test this out!**

Check out [CONTRIBUTING.md](https://github.com/th0r/webpack-bundle-analyzer/blob/master/CONTRIBUTING.md) on how to test-drive a development version of this plugin.

Just run it from my fork instead:

```
git clone git://github.com/valscion/webpack-bundle-analyzer.git
git checkout support-array-concat-bundles
```

Confirmed to work:

- [x] @matteoantoci https://github.com/th0r/webpack-bundle-analyzer/issues/29#issuecomment-277523357
- [x] @Koslun https://github.com/th0r/webpack-bundle-analyzer/issues/29#issuecomment-277669582
- [ ] @jovinbm 
- [x] @InSuperposition https://github.com/th0r/webpack-bundle-analyzer/issues/29#issuecomment-277892333
- [x] @frankleng 
- [ ] @christianalfoni 
- [ ] @eshion 
- [x] @unsafecode 
- [ ] @drcmda

PR #45, opened 2017-02-08
Support modules defined in Array CallExpressions

jharris4:
I just created a fork to add support for extra bundles with modules defined by `Array(N).concat([])`.

I really wish I had noticed the existing pull request https://github.com/th0r/webpack-bundle-analyzer/pull/44 before I did this work!

My solution looks very similar to the existing PR, but I figured I'd make a PR anyway since I'd already done the work.

For what it's worth, I've tested my fork against a couple of large Webpack projects where we have extra bundles with the Array CallExpressions, and it seems to be working great. I guess this probably validates the existing PR as well, since our implementations are so similar.


th0r:
@jharris4 Thanks a lot for your work but I does the same as #44.

PR #48, opened 2017-02-12
Ability to set host address where HTTP server should listen

freaz:
Allows to set different bind host for Analyzer server.

Example use is when Analyzer is running inside Docker container with Virtualbox.
This setup requires to access running applications on different ip than localhost. So we need have way how make Analyzer server listen on `http://192.168.0.99:8888` or similar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ new BundleAnalyzerPlugin({
   // In `static` mode single HTML file with bundle report will be generated.
   // In `disabled` mode you can use this plugin to just generate Webpack Stats JSON file by setting `generateStatsFile` to `true`.
   analyzerMode: 'server',
+  // Host that will be used in `server` mode to start HTTP server.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -44,6 +44,8 @@ new BundleAnalyzerPlugin({
   // In `static` mode single HTML file with bundle report will be generated.
   // In `disabled` mode you can use this plugin to just generate Webpack Stats JSON file by setting `generateStatsFile` to `true`.
   analyzerMode: 'server',
+  // Host that will be ised in `server` mode to start HTTP server.
```
valscion: Probably should be `used` instead of `ised`

valscion:
Seems like a good and easy feature to me.

README.md
freaz: Thanks fixed.

th0r:
Thanks, merged!

PR #50, opened 2017-02-13
Don't add a module to a non-module object

### LINKED ISSUE #49: Incompatible with webpack 2
#### Issue description

Was working brilliantly till I attempted to update to WebPack 2.2.0

#### Technical info
* Webpack Bundle Analyzer version: 2.2.3
* Webpack version: 2.2.0
* Node.js version: 6.9.3
* npm/yarn version: 0.18.1 (yarn)
* OS: macOS

#### Debug info

Tried both, same result

No options, just the defaults

Can't generate stats.json as it blows up before then.

#### Error

```
Could't analyze webpack bundle:
TypeError: currentFolder.addModule is not a function
Could't find any javascript bundles in provided stats file
```

MarshallOfSound:
Fixes #49  

Basically after some logging it looked like `currentFolder` in some situations ended up being a module, and modules don't have a `addModule` method.  This "fixed" it.

Not too familiar with the codebase so not sure if this could potentially break something

### DIFF
--- a/src/tree.js
+++ b/src/tree.js
@@ -160,6 +160,7 @@ class Folder extends Node {
       currentFolder = currentFolder.getChild(folderName) || currentFolder.addFolder(folderName);
     });
 
+    if (!currentModule.addModule) return;
     currentFolder.addModule(fileName, data);
   }
 
--- a/src/tree.js
+++ b/src/tree.js
@@ -160,7 +160,7 @@ class Folder extends Node {
       currentFolder = currentFolder.getChild(folderName) || currentFolder.addFolder(folderName);
     });
 
-    if (!currentModule.addModule) return;
+    if (!currentFolder.addModule) return;
     currentFolder.addModule(fileName, data);
   }
 


src/tree.js
```diff
@@ -160,6 +160,7 @@ class Folder extends Node {
       currentFolder = currentFolder.getChild(folderName) || currentFolder.addFolder(folderName);
     });
 
+    if (!currentModule.addModule) return;
```
valscion: What's `currentModule` here? Should this be `currentFolder` instead?

valscion:
I don't know the history of the checks above your change, but it's likely that they try to avoid the exact situation your code crashes in:

```js
   addModuleByPath(path, data) {
     const [folderNames, fileName] = [path.slice(0, -1), _.last(path)];
     let currentFolder = this;
 
     _.each(folderNames, folderName => {
        // vvvvvv This line below, maybe that doesn't work correctly for you for some reason?
        currentFolder = currentFolder.getChild(folderName) || currentFolder.addFolder(folderName);
      });
```

If you'd be able to provide a reproduction repository, we might be able to create a test case for this change. Otherwise we'd risk breaking it with future changes.

MarshallOfSound:
@valscian I'll try create a small example if I get time, but this seems to be an edge case with webpacks new bundling system so not sure if I'll actually be able to create a reproduction case.

Looking in at the code now instead of hacking around it it appears that

```js
currentFolder.getChild(folderName)
```

Could technically return a module and not a folder?  (At least that's what appears to be happening)

src/tree.js
MarshallOfSound: Woops...  That's awkward

valscion:
I'd very much love to know _what_ this exactly fixes, i.e. find the root cause instead of treat the symptoms.

I don't have time myself to create a reproduction when I don't even have a failure in my own code in the first place. I'd be extremely grateful if you could provide a reproduction so that I could help with coming up a good test case.

Without a new test case, I would be too afraid to accept this as we will most likely break this in the future by accident. Sorry :/

th0r:
> Could technically return a module and not a folder?

I think that's exactly what happens but we should understand why...

damienadermann:
👍  This fixes a problem I'm having using https://github.com/thejameskyle/react-loadable. I can't see what the react-loadable package is doing different that is causing this.

damienadermann:
To add to this, I just had a play around I am using webpacks import in the same module that imports the react-loadable. Including just react-loadable *doesn't* cause the error even though thats the module than can't be imported. Its not until I add the `import` call that it breaks.

valscion:
Ok, so this might happen because of a chunk entry point? That would make sense. It would be nice if you could create a small reproduction repository that has this error, so I could fiddle around it and figure out why this happens.

relekang:
@damienadermann This problem disappeared in my project when I updated react-loadable to 3.0.1. Hope it helps :v:

valscion:
Closing this PR as #62 fixed the underlying reason. Thank you for your contributions!

PR #56, opened 2017-03-04
Fix parsing JSONP chunks with entry points for webpack 2

### LINKED ISSUE #54: Fails to parse bundle asset.
#### Issue description
I am using the analyser as a plugin to get the gzipped stats.
Recently some code change caused this to happen.

Couldn't parse bundle asset "main-1468dcadd5cb43aed7c5.js".
Analyzer will use module sizes from stats file.

#### Technical info
* Webpack Bundle Analyzer version: 2.3.0
* Webpack version: 2.2.1
* Node.js version: 6.9.2
* npm/yarn version: 4.1.2
* OS: OSX 10.11

#### Debug info

As a plugin.

Stats and bundles: https://www.dropbox.com/s/9kbr4lyljlsvqgw/stats%20and%20bundle.zip?dl=0

cc @valscion

valscion:
If the JSONP chunk contains modules used as entry points, webpack 2 will
tack the entry point module IDs as the third argument to the JSONP
function.

https://github.com/webpack/webpack/blob/v2.2.1/lib/JsonpChunkTemplatePlugin.js#L19-L21

This fixes #54, at least given the same input I now can generate a valid report.


valscion:
@ruiaraujo, can you confirm that this fixes your issue?

Check out [CONTRIBUTING.md](https://github.com/th0r/webpack-bundle-analyzer/blob/master/CONTRIBUTING.md) on how to test-drive a development version of this plugin.

Just run it from my fork instead:

```
git clone git://github.com/valscion/webpack-bundle-analyzer.git
git checkout support-chunks-with-entries
```

ruiaraujo:
On holidays until Tuesday with no laptop. But when I  am back, I will take it for a test run.

ruiaraujo:
@valscion I can confirm it does work.

src/parseUtils.js
```diff
@@ -45,7 +45,7 @@ function parseBundle(bundlePath) {
         // As function name may be changed with `output.jsonpFunction` option we can't rely on it's default name.
         if (
           node.callee.type === 'Identifier' &&
-          args.length === 2 &&
+          (args.length === 2 || args.length === 3) &&
           isArgumentContainsChunkIds(args[0]) &&
           isArgumentArrayConcatContainingChunks(args[1])
```
th0r: Could you also check the structure of the third argument? Is it an array of chunk ids?

valscion: Mm yeah, I can do that. That sounds like a good protection idea.

Do you want me to add the same protection to the plain old `webpackJsonp([<chunks>], <modules>, ...)` call [is before this if-statement](https://github.com/valscion/webpack-bundle-analyzer/blob/7be38bf73fd465859a96464a91e6b9c24dc9ff2b/src/parseUtils.js#L34)?

I could do that in another PR, though, to make this PR more focused :)

th0r: Ah, we have just `args.length >= 2` there...

Now I'm thinking about do we really need to be that strict or `args.length === 2 || args.length === 3` is good enough.

What do you think?

th0r: I mean do we know for sure that this arguments contains chunk ids? If so then we could add this extra check, but in other case I would use just `length` check.

valscion: I don't see any reason why there would be anything different than chunk IDs in the third place. I think I tracked pretty well the code path in webpack that generates the output, and I didn't see any signs of other code being sent as the third argument.

It used to always even be just two args, back in webpack 1, but webpack 2 changed it to fix some bug with entry points not being assigned to globals with UMD or something.

I feel pretty safe just checking for the arguments length. If someone later has a bug with this, it should be quite easy to fix ;)

th0r:
Released in `v2.3.1`. Thanks @valscion!

PR #69, opened 2017-05-03
Add server host name option to CLI

difelice:
Adds the option to customize the _host_ when running CLI in _server_ mode.

The `host` option defaults to `0.0.0.0`.

### DIFF
--- a/src/bin/analyzer.js
+++ b/src/bin/analyzer.js
@@ -30,6 +30,12 @@ const program = commander
     'server'
   )
   .option(
+    '-h, --host <host>',
+    'Host that will be used in `server` mode to start HTTP server.' +
+    br('Default is `127.0.0.1`.'),
+    '127.0.0.1'
+  )
+  .option(
     '-p, --port <n>',
     'Port that will be used in `server` mode to start HTTP server.' +
     br('Default is 8888.'),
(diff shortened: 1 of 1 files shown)


valscion:
Ah yes, this new option from #48  never made it to the CLI. Thanks for adding it! :+1:

src/bin/analyzer.js
```diff
@@ -58,6 +65,7 @@ let {
 
 if (!bundleStatsFile) showHelp('Provide path to Webpack Stats file as first argument');
 if (mode !== 'server' && mode !== 'static') showHelp('Invalid mode. Should be either `server` or `static`.');
+if (mode === 'server' && host === '') showHelp('Invalid host name');
```
th0r: Let's use `!host` instead of `host === ''`

src/bin/analyzer.js
```diff
@@ -30,6 +30,12 @@ const program = commander
     'server'
   )
   .option(
+    '-h, --host <host>',
+    'Host that will be used in `server` mode to start HTTP server.' +
+    br('Default is 0.0.0.0.'),
```
th0r: Let's use `127.0.0.1` as the default address to match `analyzerHost` option.

difelice: Updated. Thanks.

th0r:
Merged, thanks!

th0r:
Published in `v2.5.0`

PR #82, opened 2017-05-27
Split reporter to new package

valscion:
Follow-up to #81 and #80, this contains the last missing pieces of removing the treemap UI code and using https://github.com/webpack-bundle-analyzer/reporter-treemap instead.

This should be safe to even release with a minor version bump, as the treemap code extraction is not yet visible to outside world — apart from one new package being installed with the next `webpack-bundle-analyzer` version.

I've invited you @th0r to the @webpack-bundle-analyzer GitHub organization and the [`@webpack-bundle-analyzer` npm organization](https://www.npmjs.com/org/webpack-bundle-analyzer), where I extracted the reporter to [`@webpack-bundle-analyzer/reporter-treemap`](https://www.npmjs.com/package/@webpack-bundle-analyzer/reporter-treemap) package.


valscion:
I tested this by running `npm pack` on the latest commit (95bf3c1568f33513323a5c60cf7cfdb84eedebb2), moving the packed `.tgz` file to our project folder and then running `npm install ./webpack-bundle-analyzer-2.8.2.tgz` from there. Everything is working smooooooothly ✌️

valscion:
Hmm seems like the `analyzerMode: 'server'` doesn't work for some reason, the process justs quits before it opens the browser. Looking into this now...

valscion:
I'm gonna close this PR as a proof of concept and then figure out a Lerna approach where we can keep the reporter package under the same repository :)

PR #83, opened 2017-05-29
Add typescript typings

marvinhagemeister:
This PR adds typings for TypeScript, to make it easier to use `webpack-bundle-analyzer` with it. The package `@types/webpack` is necessary, because we depend on `webpack`. Although webpack itself is listed under devDependencies, the `@types/webpack` package needs to be listed under normal `dependecies` because our typings depend on them and typings are resolved at runtime.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,45 @@
+declare module "webpack-bundle-analyzer" {
+  import { Plugin, Options } from "webpack";
+
+  export type AnalyzerMode = "server" | "static" | "disabled";
+  export type ModuleSize = "stat" | "parsed" | "gzip";
+  export type LogLevel = "info" | "warn" | "error" | "silent";
+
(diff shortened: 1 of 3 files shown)


valscion:
I don't know whether we want to maintain these typings if we change stuff. It should not be that difficult, though, but we don't have any automated way of seeing whether we broke these in some version.

Have you considered sending these typings just to https://github.com/DefinitelyTyped/DefinitelyTyped ? If not, why? I'm not that familiar with how package typings are normally handled in the TypeScript ecosystem.

marvinhagemeister:
Fair point. I'll upstream the typings to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped).

valscion:
Thanks! I'll keep this in mind if at some later point we want to have these :)

PR #96, opened 2017-07-13
Support parsing of ESNext features

### LINKED ISSUE #94: Can't parse bundle asset with babel-preset-env on latest browsers
#### Issue description
I'm getting the "Can't parse bundle asset" error (and therefore am only seeing stat sizes) when I adjust my `babel-preset-env` parameters to only support the latest browser versions (specifically, Safari 10.1). The big change here (compared to Safari 10.0, which works as expected) is that babel is using native async/await.

Bizarrely, the parse _does succeed_ on 10.1 if I set devtool to 'eval-source-map'. If I set it to null, or 'source-map', I get this error. Not entirely sure why this would be the case, but happy to throw additional logs in to help better debug...

#### Technical info
* Webpack Bundle Analyzer version: `2.8.2`
* Webpack version: `3.1.0`
* Node.js version: `7.10.0`
* npm/yarn version: `4.2.0`
* OS: `MacOS 10.12.4`

#### Debug info
I use this module as a plugin: 
```js
new BundleAnalyzerPlugin({
        analyzerMode: 'server',
	reportFilename: `${logDirectory}/statistics.html`,
	analyzerHost: 'localhost',
	analyzerPort: 9081,
	defaultSizes: 'gzip',
	openAnalyzer: true,
	generateStatsFile: true,
	logLevel: 'info',
        statsOptions: { source: false },
})
```

I'm able to reproduce this without any additional plugins. I use `babel-loader` with the following options:
```js
{
	test: /\.jsx?$/,
	exclude: /node_modules/,
	loader: 'babel-loader',
	query: {
		presets: [
			['env', {
				targets: {
					safari: '10.1',   // change to '10.0' and bundle parse succeeds
					edge: '15',
					firefox: '54',
					chrome: '60',
				},
			        modules: false,
			        loose: true,
		        }],
		],
		plugins: [
			'syntax-dynamic-import',
			'transform-class-properties',
			'transform-object-rest-spread',
			'transform-react-jsx',
		],
	},
}
```

[stats.json.txt](https://github.com/th0r/webpack-bundle-analyzer/files/1136848/stats.json.txt)

th0r:
Fixes #94

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "views"
   ],
   "dependencies": {
-    "acorn": "^5.0.3",
+    "acorn": "^5.1.1",
     "chalk": "^1.1.3",
     "commander": "^2.9.0",
     "ejs": "^2.5.6",
--- a/src/parseUtils.js
+++ b/src/parseUtils.js
@@ -9,7 +9,13 @@ module.exports = {
 
 function parseBundle(bundlePath) {
   const content = fs.readFileSync(bundlePath, 'utf8');
-  const ast = acorn.parse(content, { sourceType: 'script' });
+  const ast = acorn.parse(content, {
+    sourceType: 'script',
+    // I believe in a bright future of ECMAScript!
+    // Actually, it's set to `2050` to support the latest ECMAScript version that currently exists.
+    // Seems like `acorn` supports such weird option value.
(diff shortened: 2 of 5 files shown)


valscion:
Nice :+1:

PR #97, opened 2017-07-14
Version 3: Split reporting UI from dependency graph size calculation

valscion:
This is a tracking PR for getting to version 3.

## What? What is version 3?

Version 3 will split the reporting UI (the treemap reporter) from the webpack code.

This allows users to create any kind of reporter, while the webpack plugin and CLI would provide the raw data containing post-minification and gzipped sizes.

## How can you help?

Please leave me a comment on your thoughts on this idea, or even add a reaction to this PR description so I could see how many are interested in this at all.

## How did I come up with this idea?

The thing that started this eventual process was an experimental PR #25 — I wanted to see if I could add a sunburst chart visualizer to `webpack-bundle-analyzer` alongside the treemap visualizer.

It seems that I wasn't the only one, @bitpshr had also done [a much more finished version of a sunburst chart](https://github.com/th0r/webpack-bundle-analyzer/pull/25#issuecomment-265478690):

> Related, I [added sunburst support as well](https://github.com/th0r/webpack-bundle-analyzer/compare/master...bitpshr:sunburst-chart?expand=1) (I didn't know @valscion's PR existed).
>
> @th0r, you built a great tool and we plan to use it as a basis for Dojo 2 build reporting. We were concerned about shipping the Carrot Search commercial treemap solution to users with permanent demo branding, hence our fork.
>
> sunburst

And sunbursts are just one possible visualizer. What about a VR-based bundle content visualizer? https://twitter.com/ken_wheeler/status/872939627656728578

## What has happened before this PR?

I have figured out how `webpack-bundle-analyzer` works and pondered what would be the best approach to have a custom reporter.

And then me and @th0r figured that the best solution would be to split analyzer from the treemap UI code: https://github.com/th0r/webpack-bundle-analyzer/issues/32

I've tested various different approaches to this problem: https://github.com/th0r/webpack-bundle-analyzer/pull/40 (with explanation in https://github.com/th0r/webpack-bundle-analyzer/issues/32#issuecomment-274270668), a proof of concept of this in https://github.com/valscion/webpack-bundle-analyzer/pull/1 and generally just figuring things out.

## Architectural changes

This repository is converted to a monorepo (like [Babel](https://github.com/babel/babel/blob/master/doc/design/monorepo.md)), and the maintenance work is handled with [Lerna](https://github.com/lerna/lerna/).

New packages (to be) created:

- [x] webpack plugin code
- [x] Treemap reporter that uses the Carrot Search Foamtree
- [x] Bundle content parser
- [x] Logger
- [ ] Compatibility package to tie these together exactly like existing `webpack-bundle-analyzer` plugin works to help in migration

These package splits are still not set in stone, and I'll evaluate the best choices for split points as I go.

## TODO (that will change as I think things through)

I intend to work on this by creating new PRs for distinct features that point to this branch instead of `master`. This way I can work in smaller steps, and hopefully have more focused PRs.

- [x] Merge in a refactoring that moved chart data calculation to source code edges, #81
- [x] Proper scripts for the common development cases
- [x] Run `build` automatically for all packages before publishing new versions
- [x] Move all reporter related configuration options under reporter specific options object
- ~~Improve the current test infrastructure~~~
  * Postponed to a later time
- [ ] Create a new package that has exactly same API as v2 does, but just utilizes these new packages
  * This should make migration cost zero _and_ still allow using a different visualizer by just using a different package as the entry point.

Nice things to have:

- [ ] Introduce [Flow](https://github.com/facebook/flow), at least to the `chartData` boundaries, and use the same types in reporter and plugin
- [ ] Maybe add some way of validating `reporterOptions` which reporters themselves can use


valscion:
Ach, it's way too slow to figure out distinct PRs for all the changes 😅 . I'm gonna just commit directly here, and try to report some comments in between as to what I've done.

.travis.yml
```diff
@@ -3,6 +3,7 @@ sudo: false
 branches:
   only:
     - master
+    - version-3
```
valscion: This will need to be removed _right before merging_. This is needed here so that branches against this `version-3` branch have their tests ran in CI. See https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/143#issuecomment-362899898

valscion:
I think I might get to this again at some point, but this branch has lived for long enough so far that continuing from here would cause undue hassle. I'll close this PR for now.

PR #103, opened 2017-07-17
Improve testing infra, switch to Jest ran on root-level

valscion:
This PR focuses on splitting out the integration level tests from current plugin code to separate, root-level `integration_tests` directory.

Jest is a more robust and painless test runner than Mocha so I convert all tests to Jest if at all possible. This is just so that I could have fun while doing tests.

Part of #97


valscion:
It seems that I got stuck on this, so I'm gonna abandon these changes and maybe look into shaving this yak later :smile:

PR #108, opened 2017-07-30
Allow webpack modules to be wrapped in arrow functions.

regiontog:
When using e.g. [uglifyjs](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) with options ```{uglifyOptions: { ecma: 8 }}``` bundles with arrow functions are generated. This caused the analyser to always use ```stats``` over ```parsed``` and ```gzipped``` because it could not parse the bundles.

I don't know what's necessary to properly support this, but with this fix my project works nicely.


th0r:
Thanks a lot!

Issue #113, opened 2017-09-06
Add brotli size support


aldo-roman:
This is a discussion issue to add Brotli size support on the report, side by side with gzip.
On the app we support, 78% of users surf the web with a brotli-enabled browser.
Global support is around 60%.
See http://caniuse.com/#search=brotli

Also, in our tests, Brotli grants 10% more compression than Gzip.
We are developing a Angular CLI, Angular 4+ app.

I will be experimenting with a fork to add Brotli by replicating the Gzip size implementation.
Any suggestion and feedback is welcome.

valscion:
Huh, never heard of Brotli before 😄 . Awesome to learn new things!

@th0r should be able to answer with any suggestions and feedback, when he's got time. I'd like to at least see a proof-of-concept of this Brotli support if possible ☺️

th0r:
Never heard of Brotli as well. The only concern I have is the increasing of analyze time. Agree with @valscion here - it would be nice to see a proof-of-concept PR and comparison of analyze times with and without it.

valscion:
It might be possible to add brotli support if it would be calculated only when some CLI or plugin option would be there. That way it would not increase analyze time for those who don't yet care about Brotli.

th0r:
👍

aldo-roman:
Hi guys, here there is the PR for Brotli support. It is working without problems on my setup. I couldn't  include an opt-in option for the feature.

Here is a screenshot on how this looks:
![image](https://user-images.githubusercontent.com/497197/30229436-8e0be2b8-94a7-11e7-9d04-590216fc6207.png)

valscion:
Neat! I think it'll be necessary to have an opt-in somehow, but currently it might require some architectural changes. Large changes would cause difficult merge conflicts with #97, though — I hope I could get that finished some day 😕

I hope you're OK with using the Brotli support from your branch? It's a bit unlikely that we'll want to calculate these sizes by default as brotli isn't that widely used yet

edmorley:
If this does get implemented, please can the brotli related dependencies not be added to `dependencies` - to avoid issues like those in siddharthkp/bundlesize#202. (eg check for the dependencies at runtime and if not found and "brotli mode" enabled, output a suitable message.)

[...]

PR #114, opened 2017-09-08
Add support for Brotli size

aldo-roman:
Solves #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ This module will help you:
 4. Optimize it!
 
 And the best thing is it supports minified bundles! It parses them to get real size of bundled modules.
-And it also shows their gzipped sizes!
+And it also shows their gzip and brotli sizes!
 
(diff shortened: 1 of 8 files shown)


RinkAttendant6:
@valscion Any plans on getting this feature in?

valscion:
@RinkAttendant6 not as it is right now. #113 contains more context. The crux is that we don't want this to increase analyze time, and this feature would have to be opt-in. Currently it might require some architectural changes, though, so it isn't that easy to do.

valscion:
Hi @aldo-roman! Thank you for your contributions. Displaying Brotli size would be definitely useful for some folks.

However, due to the reasons listed in #113, I'm afraid we can't go on with this PR as it is now. I hope we'll get this library's architecture to a point where adding optionality for new sizes would be simpler. Right now we aren't there 😞

PR #135, opened 2017-12-27
Update README.md to match webpack-contrib style

valscion:
I took https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/v3.0.2/README.md as an example of the README.md structure.

See the rendered version here: https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/readme-webpack-contrib/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,10 +89,34 @@ webpack-bundle-analyzer bundle/output/path/stats.json
 
 <h2 align="center">Options (for CLI)</h2>
 
-View available options by running the CLI with `--help` flag:
+```
+Usage: webpack-bundle-analyzer <bundleStatsFile> [bundleDir] [options]
 
-```bash
-webpack-bundle-analyzer --help
+Arguments:
+
+  bundleStatsFile  Path to Webpack Stats JSON file.
+  bundleDir        Directory containing all generated bundles.
+                   You should provided it if you want analyzer to show you the real parsed module sizes.
+                   By default a directory of stats file is used.
+
+Options:
+
+  -h, --help                  output usage information
+  -V, --version               output the version number
+  -m, --mode <mode>           Analyzer mode. Should be `server` or `static`.
+                              In `server` mode analyzer will start HTTP server to show bundle report.
+                              In `static` mode single HTML file with bundle report will be generated.
+                              Default is `server`.
+  -h, --host <host>           Host that will be used in `server` mode to start HTTP server.
+                              Default is `127.0.0.1`.
+  -p, --port <n>              Port that will be used in `server` mode to start HTTP server.
+                              Default is 8888.
+  -r, --report <file>         Path to bundle report file that will be generated in `static` mode.
+                              Default is `report.html`.
+  -s, --default-sizes <type>  Module sizes to show in treemap by default.
+                              Possible values: stat, parsed, gzip
+                              Default is `parsed`.
+  -O, --no-open               Don't open report in default browser automatically.
 ```
 
 <h2 align="center" id="size-definitions">Size definitions</h2>
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ as Uglify, then this value will reflect the minified size of your code.
 
 ### `gzip`
 
-This is the size of running the file(s) through gzip compression.
+This is the size of running the parsed bundles/modules through gzip compression.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,66 +45,54 @@ This module will help you:
...
+This is the "output" size of your files. If you're using a Webpack plugin such
+as Uglify, then this value will reflect the minified size of your code.
 
-### As a CLI utility
-You can also analyze already existing bundles if you have a Webpack Stats JSON file.
+### `gzip`
+
+This is the size of running the file(s) through gzip compression.
+
+<h2 align="center">Usage as a CLI utility</h2>
```
michael-ciniawsky: I personally would move this under  `Usage` as 

```
### `CLI`

...Contents
```

### `CLI`

...Contents

valscion: Good idea, I'll have to try it out to see how it feels :)

What about the rest of the changes, any other details that pop up?

README.md
```diff
@@ -18,66 +45,54 @@ This module will help you:
...
+|**`analyzerPort`**|`{Number}`|Default: `8888`. Port that will be used in `server` mode to start HTTP server.|
+|**`reportFilename`**|`{String}`|Default: `report.html`. Path to bundle report file that will be generated in `static` mode. Relative to bundle output directory (which is `output.path` in webpack config).|
+|**`defaultSizes`**|One of: `stat`, `parsed`, `gzip`|Default: `parsed`. Module sizes to show in report by default. See explanation below this table.|
+|**`openAnalyzer`**|`{Boolean}`|Default: `true`. Automatically open report in default browser.|
+|**`generateStatsFile`**|`{Boolean}`|Default: `false`. If `true`, webpack stats JSON file will be generated in bundle output directory|
+|**`statsFilename`**|`{String}`|Default: `stats.json`. Name of webpack stats JSON file that will be generated if `generateStatsFile` is `true`. Relative to bundle output directory.|
+|**`statsOptions`**|`null` or `{Object}`|Default: `null`. Options for `stats.toJson()` method. For example you can exclude sources of your modules from stats file with `source: false` option. See more options here: https://github.com/webpack/webpack/blob/webpack-1/lib/Stats.js#L21 |
```
th0r: It's better to actualize a link to the sources.

README.md
```diff
@@ -18,66 +45,54 @@ This module will help you:
...
+### `parsed`
+
+This is the "output" size of your files. If you're using a Webpack plugin such
+as Uglify, then this value will reflect the minified size of your code.
 
-### As a CLI utility
-You can also analyze already existing bundles if you have a Webpack Stats JSON file.
+### `gzip`
+
+This is the size of running the file(s) through gzip compression.
```
th0r: Let's change it to something like this: `This is the size of running the parsed bundles/modules through gzip compression.`

th0r:
Thanks @valscion. Added a few small comments.

valscion:
Thanks for the comments @th0r and @michael-ciniawsky 😊. I'll update this PR later today to address both of your comments.

valscion:
How does it look like now, @th0r?

I'll do another PR for further improvements, like documenting the CLI options (maybe even auto-generate them somehow) and maybe coming up with a better link to `stats.toJson()` options than webpack@1 source code.

th0r:
I liked previous structure more. Now it's a little confusing because of the sections order:
- Usage in code
- Usage in CLI
- Options (for usage in code)
- Options for CLI
This switching of context is IMHO a little confusing.

I liked previous structure more:
- Usage in code
- Options for usage in code
- Usage in CLI
- Options for usage in CLI

valscion:
Yeah makes sense. I agree with you.

valscion:
Ok I reorganized the readme — it makes sense with the structure you proposed @th0r.

The CLI options part seemed redundant and empty, so I filled it in with mostly copy-paste from what `--help` outputs. I checked out [commander-to-markdown](https://github.com/studio-b12/commander-to-markdown) package, but the way the `src/bin/analyzer.js` is written wasn't straightforward to appease that tool.

valscion:
Is this good to go? If we'd want to tweak this further, I'd rather do it in a separate pull request unless there are big objections against this direction. I'd love it if this PR didn't stall for too long :)

th0r:
Yep, everything is great! Merged!

Issue #138, opened 2018-01-02
Server errors are unhandled


sompylasar:
#### Issue description

After introducing `webpack-bundle-analyzer`, the following two errors started crashing the development server process.

When accidentally using the same `analyzerPort` for the plugin in two webpack configs within the same process (universal, for browser-side and server-side) – configuration issue, has a workaround:
```
Unhandled exception:  Error: listen EADDRINUSE 127.0.0.1:8888
     at Object._errnoException (util.js:1021:11)
     at _exceptionWithHostPort (util.js:1043:20)
     at Server.setupListenHandle [as _listen2] (net.js:1344:14)
     at listenInCluster (net.js:1385:12)
     at doListen (net.js:1494:7)
     at _combinedTickCallback (internal/process/next_tick.js:141:11)
     at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```

When quickly refreshing the analyzer webpage while it's still loading – no workaround possible:
```
Unhandled exception:  Error: read ECONNRESET
     at _errnoException (util.js:1021:11)
     at TCP.onread (net.js:608:25)
```

#### Technical info
* Webpack Bundle Analyzer version: `webpack-bundle-analyzer@2.9.1`
* Webpack version: `webpack@2.7.0`
* Node.js version: `v8.7.0`
* npm/yarn version:  `npm@5.6.0`
* OS: `Darwin ... 16.0.0 Darwin Kernel Version 16.0.0: Mon Aug 29 17:56:20 PDT 2016; root:xnu-3789.1.32~3/RELEASE_X86_64 x86_64`

#### Debug info

> How do you use this module? As CLI utility or as plugin?

As a plugin.

> If CLI, what command was used? (e.g. `webpack-bundle-analyzer -O path/to/stats.json`)

N/A

> If plugin, what options were provided? (e.g. `new BundleAnalyzerPlugin({ analyzerMode: 'disabled', generateStatsFile: true })`)

```js
    new BundleAnalyzerPlugin({
      analyzerMode: 'server',
      analyzerHost: '127.0.0.1',
      analyzerPort: (CLIENT ? 8888 : 8889),  // This conditional port fixes EADDRINUSE.
      openAnalyzer: false,
      generateStatsFile: false,
    }),
```

> What other Webpack plugins were used?

Oh, many of them, via a dynamically generated webpack config. I cannot share that.

> It would be nice to also attach webpack stats file.
> It can be generated using these options:
> ```js
[...]

PR #140, opened 2018-01-06
Add a listener for the 'error' event on the WebSocket server client

lpinca:
This prevents the process from exiting when an `'error'` event is emitted on the WebSocket server client.

Refs: https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/138

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+sudo: false
 branches:
   only:
     - master
@@ -10,21 +11,12 @@ cache:
   - yarn: true
   - directories:
     - node_modules
-env:
-  # Required to compile native modules for Node.js v4+
-  # See https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements
-  - CXX=g++-4.8
 addons:
   apt:
-    sources:
-      # Required to compile native modules for Node.js v4+
-      - ubuntu-toolchain-r-test
     packages:
       # Required by Electron
       # https://github.com/juliangruber/electron-stream#travis
       - xvfb
-      # Required to compile native modules for Node.js v4+
-      - g++-4.8
 before_script:
   # Required by Electron
   - export DISPLAY=':99.0'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@
 
 _Note: Gaps between patch versions are faulty, broken or test releases._
 
+## UNRELEASED
+
+ * **Bug Fix**
+   * Add a listener for the 'error' event on the WebSocket server client (#140)
+
+ * **Internal**
+   * Clean up .travis.yml (#140)
+   * Update ws to version 4.0.0 (#140)
+
 ## 2.9.1
 
  * **Bug Fix**
@@ -18,7 +27,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 ## 2.9.0
  * **New Feature**
    * Show chunk sizes in sidebar (closes #91)
-   
+
  * **Bug Fix**
    * Properly parse webpack bundles that use arrow functions as module wrappers (#108, @regiontog)
 
@@ -31,7 +40,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 ## 2.8.2
  * **Improvement**
    * Greatly improved accuracy of gzip sizes
- 
+
  * **Bug Fix**
    * Generate report file in the bundle output directory when used with Webpack Dev Server (fixes #75)
 
@@ -43,7 +52,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
  * **Improvement**
    * Analyzer now supports `webpack --watch` and Webpack Dev Server!
      It will automatically update modules treemap according to changes in the sources via WebSockets!
- 
+
  * **Internal**
    * Use `babel-preset-env` and two different Babel configs to compile node and browser code
    * Update deps
@@ -67,7 +76,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 ## 2.4.0
  * **Bug Fix**
    * Fix `TypeError: currentFolder.addModule is not a function`
-  
+
  * **Internal**
    * Update deps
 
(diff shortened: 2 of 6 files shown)


valscion:
This looks great! Is there a way I could verify that this fix works? Even a manual test would be amazing.

lpinca:
Manual test is easy, use Chrome establish a WebSocket connection and refresh the page as per https://github.com/websockets/ws/issues/1256.

For a unit test it depends, a WebSocket protocol error is easy to test, just send an invalid frame. For a network error is a bit more complex but still possible, see https://github.com/websockets/ws/blob/master/test/websocket.test.js#L1198-L1227 for example.

I'm not familiar with this project but I can try adding a unit test for a protocol error.

valscion:
I don't happen to have any project that utilizes the dev server with webpack-bundle-analyzer, so that's why it's a bit difficult for me to make sure this resolves any issue. It would be amazing if you'd have the energy and time to create a small reproduction repository where I could test this out.

It's been one goal of mine to eventually get this project to a state where we'd at least have manual reproduction cases available for quickly testing stuff like this. Would you be interested in helping out with making this a reality?

.travis.yml
```diff
@@ -10,21 +11,12 @@ cache:
   - yarn: true
   - directories:
     - node_modules
-env:
-  # Required to compile native modules for Node.js v4+
-  # See https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements
-  - CXX=g++-4.8
```
lpinca: Ubuntu Trusty is now the default on Travis CI so this is no longer required as the gcc version is 4.8.4

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: node_js
+sudo: false
```
lpinca: This is to use the container based infrastructure. The start time is a lot faster.

lpinca:
Hmm weird, the test works locally but not on travis. I'll spin up an Ubuntu VM later and debug it.

valscion:
You could try increasing the timeout from the default 2000 milliseconds. For example, the dev server tests use a timeout of 15000ms: https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/v2.9.0/test/dev-server.js

lpinca:
Ok found the issue, the `ws` version is pinned to 3.3.1 and the test uses behavior of `ws>=3.3.3`.

lpinca:
This is ready for me, if there is anything you don't like or want changed/removed let me know.

test/viewer.js
```diff
@@ -0,0 +1,50 @@
...
+          '\r\n'
+        ].join('\r\n'));
+      });
+
+      socket.on('data', function (chunk) {
+        const expected = Buffer.from([
+          'HTTP/1.1 101 Switching Protocols',
+          'Upgrade: websocket',
+          'Connection: Upgrade',
+          'Sec-WebSocket-Accept: XYMa8DJmsxHNEsGJZwEZT7XX1Dk=',
```
valscion: Could we generate this response using the `crypto` module instead of hard-coding it? I didn't know about these headers, and it would've been nice to know where this magical looking value came from.

```js
const crypto = require('crypto');
const websocketAcceptHash = crypto.createHash('sha1');
websocketAcceptHash.update(
  // The client-generated "Sec-WebSocket-Key" value
  'Ix02x1E8kmC9ie01/dGCuw==' +
  // The GUID constant defined in WebSocket protocol draft
  // https://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-17#section-1.3
  '258EAFA5-E914-47DA-95CA-C5AB0DC85B11'
);
const websocketAccept = hash.digest('base64');
```

test/viewer.js
```diff
@@ -0,0 +1,50 @@
...
+          '\r\n'
+        ].join('\r\n'));
+
+        expect(chunk.equals(expected)).to.be.true;
+
+        // Send a WebSocket frame with a reserved opcode: 5
+        socket.write(Buffer.from([0x85, 0x00]));
```
valscion: Is this the part where an error is triggered? I'm not that well versed in how websocket protocol works.

If writing a reserved opcode indeed causes the error to be triggered, maybe the comment could be amended to say that explicitly? For example, something like this:

```js
// Send a WebSocket frame with a reserved opcode (5) to trigger
// an error to be emitted on the server
```

test/viewer.js
```diff
@@ -0,0 +1,50 @@
...
+    startServer(bundleStats, options).then(function ({ http: server }) {
+      const socket = net.createConnection(server.address().port, function () {
+        socket.write([
+          'GET / HTTP/1.1',
+          'Host: localhost',
+          'Upgrade: websocket',
+          'Connection: Upgrade',
+          'Sec-WebSocket-Key: Ix02x1E8kmC9ie01/dGCuw==',
+          'Sec-WebSocket-Version: 13',
+          '\r\n'
```
valscion: Do we need an additional `\r\n` here or would an empty string be enough? I see that you're joining the resulting array with `\r\n` anyway, so this seems a bit excessive.

If we indeed want to have two empty lines, could we instead have two empty strings in the array before joining?

.travis.yml
valscion: Thanks!

valscion:
Would you be open to adding a changelog about both the `ws` bump and this fix? You could add a new `## UNRELEASED` section like I've done here: https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/082bb9aabd06d9eb43df45b7e42e99086311cea8/CHANGELOG.md

test/viewer.js
lpinca: Each header line must end with `\r\n` and we need an empty one ending `\r\n` to signal the end.

test/viewer.js
lpinca: Sure.

test/viewer.js
lpinca: Yes this how the error is triggered on the server, I will update the comment.

lpinca:
@valscion PTAL.

CHANGELOG.md
```diff
@@ -18,7 +27,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 ## 2.9.0
  * **New Feature**
    * Show chunk sizes in sidebar (closes #91)
-
```
lpinca: My editor removes trailing spaces automatically. I can readd them if wanted.

valscion: It's fine :+1:

valscion:
Thank you for this! I released it in version 2.9.2 just now :)

PR #143, opened 2018-01-13
Split bundle parsing to new package

valscion:
Part of Version 3 upgrade #97, this PR aims to split the bundle parsing logic out of the plugin code to own package.

This new package could be useful outside of webpack-bundle-analyzer itself. Other tools who would like to access the parsed output would be able to do it and do whatever they want with the data.

TODO:

- [x] Split `src/analyzer.js` and related code to the new package


valscion:
What's up with TravisCI not running tests for this branch?! 😩

PR #148, opened 2018-02-01
Simplify ESLint configs

th0r:
Use my configs and extend from them.


evilebottnawi:
@th0r i think better release v3 and when migrate :+1:

th0r:
Agree!

PR #149, opened 2018-02-02
Simplify gulp tasks

th0r:
[This workaround](https://github.com/webpack-contrib/webpack-bundle-analyzer/compare/simplify-gulp-tasks?expand=1#diff-b9e12334e9eafd8341a6107dd98510c9R28) can be removed after https://github.com/gulpjs/glob-watcher/pull/34 is merged.


valscion:
Ah lovely. This will help me with v3, even a little bit, as I've had trouble deciphering how all the different parts work with gulp. Thanks for making it simpler :)

PR #159, opened 2018-03-01
Fix parsing webpack 4's chunks

jdelStrother:
Webpack 4 has a new chunk loading style, looking something like this:

```js
(window.webpackJsonp=window.webpackJsonp||[]).push([[27],{"57iH":function(e,n,t){console.log("hello world")}}]);
```

WDYT to this approach?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
+ * **Improvement**
+   * Add support for parsing Webpack 4's chunked modules ([#159](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/159), [@jdelStrother](https://github.com/jdelStrother))
+
 ## 2.11.0
 
  * **Improvement**
--- a/src/parseUtils.js
+++ b/src/parseUtils.js
@@ -72,7 +72,7 @@ function parseBundle(bundlePath) {
         }
 
         // Additional bundles with webpack 4 are loaded with:
-        // (window.webpackJsonp=window.webpackJsonp||[]).push([[minimum ID], <module>, <module>]);
+        // (window.webpackJsonp=window.webpackJsonp||[]).push([[chunkId], [<module>, <module>], [[optional_entries]]]);
         if (
           isWindowPropertyPushExpression(node) &&
           args.length === 1 &&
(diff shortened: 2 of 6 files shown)


valscion:
Hi, and welcome! Thank you for your PR :relaxed:

I was wondering, is there any case where the "minimum ID" is needed? The `[27]` part in your example. Is there a way for webpack 4 to generate numerical IDs instead of these modules keyed by a hash of some sort?

This PR in itself looks great! Would you be open to checking whether there's new sorts of chunk loading styles that also start with the `(window.webpackJsonp=window.webpackJsonp||[]).push` approach? Something similar to this:

* Extra bundle with modules inside Array concat: [bundle](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/test/bundles/validExtraBundleWithModulesInsideArrayConcat.js) -> [parsed modules](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/test/bundles/validExtraBundleWithModulesInsideArrayConcat.modules.json)
* JSONP with Array concat and entry point: [bundle](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/test/bundles/validJsonpWithArrayConcatAndEntryPoint.js) -> [parsed modules](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/test/bundles/validJsonpWithArrayConcatAndEntryPoint.modules.json)

As you can see in the above example, the module IDs are affected by the ID in the first parameter.

Figuring this out can be done outside of this PR, though — this can go in on itself.

Could you create a changelog entry to CHANGELOG.md and link back to this PR, similarly to how the links have been done in v2.10.0 and above?

jdelStrother:
> I was wondering, is there any case where the "minimum ID" is needed? The [27] part in your example. Is there a way for webpack 4 to generate numerical IDs instead of these modules keyed by a hash of some sort?

Yeah, re-reading it, the minimum id isn't necessary to calculate the ids in the chunk, so it's misleadingly named.   I'll change it.

> Would you be open to checking whether there's new sorts of chunk loading styles

The only place where I could find webpackJsonp being used is here - https://github.com/webpack/webpack/blob/master/lib/web/JsonpChunkTemplatePlugin.js.  There's an optional third argument (`entries`), which happens to work ok with my original code, but I'll add a test to confirm it.

> Could you create a changelog entry to CHANGELOG.md and link back to this PR

will do

valscion:
> The only place where I could find webpackJsonp being used is here - https://github.com/webpack/webpack/blob/master/lib/web/JsonpChunkTemplatePlugin.js. There's an optional third argument (entries), which happens to work ok with my original code, but I'll add a test to confirm it.

:+1:

I'll add a non-master link to the file in question so it won't get lost if files are moved around as webpack is developed: https://github.com/webpack/webpack/blob/v4.0.1/lib/web/JsonpChunkTemplatePlugin.js

valscion:
Looks great! Thank you for contributing! :tada:

valscion:
Released in v2.11.1 :tada:

PR #180, opened 2018-05-15
Pretty-format the generated stats.json

edmorley:
Previously the serialised stats object was all on one line, which:
* was hard for a human to read
* caused performance problems with certain text editors
* doesn't diff well

It will now be pretty-printed with 2 space indent:
https://github.com/philbooth/bfj#options-for-serialisation-functions


th0r:
Could you add a changelog entry under `UNRELEASED` section?

edmorley:
Sure! Updated :-)

Issue #160, opened 2018-03-08
Angular CLI with scripts forces displaying of only stat size


zbuhler:
#### Issue description

When using the webpack-bundle-analyzer in conjunction with the Angular CLI there is an issue with parsing the `scripts.[hash].bundle.js.`

When specifying scripts to put in the script bundle via the .angular-cli.json those scripts get concatenated and not wrapped in any webpack functionality like "`webpackJsonp(`". This appears to result in a parsing error with the webpack-bundle-analyzer which in turn causes all chunks to go off the stats.json making it so you can't see parsed or gzipped sizes when using the tool.

To reproduce this issue please use this [repo](https://github.com/zbuhler/angularCLIScriptsOnlyStatSize)

For extra information surrounding this issue (and some ideas on ways to tackle it) you can view the last couple of comments from this [isssue/thread](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/147#issuecomment-371045968).

To "fix" it locally you can do the following:

The relevant parts of the stats.json to remove are-
```diff
  "assetsByChunkName": {
    "inline": "inline.86cf029305c934704a61.bundle.js",
    "main": "main.66983cb30db00a1c5598.bundle.js",
    "polyfills": "polyfills.dc963f90580f1044b785.bundle.js",
    "styles": "styles.9c0ad738f18adc3d19ed.bundle.css",
-   "scripts": "scripts.d464be456471b1b5f048.bundle.js"
  }
```

and

```diff
-{
-  "name": "scripts.d464be456471b1b5f048.bundle.js",
-  "size": 158555,
-  "chunks": [
-    "scripts"
-  ],
-  "chunkNames": [
-    "scripts"
-  ],
-  "emitted": true
-},
```
Which effectively excludes that chunk from being analyzed.

#### Technical info
* Webpack Bundle Analyzer version: 2.11.1
* Webpack version: 3.10.0
* Node.js version: 9.1.0
* npm/yarn version: Yarn 1.3.2
* OS: Mac

#### Debug info
How do you use this module? As CLI utility or as plugin? CLI

If CLI, what command was used? (e.g. `webpack-bundle-analyzer -O path/to/stats.json`)

`webpack-bundle-analyzer dist/stats.json`

If plugin, what options were provided? (e.g. `new BundleAnalyzerPlugin({ analyzerMode: 'disabled', generateStatsFile: true })`)

N/A

#### What other Webpack plugins were used?

[...]

PR #181, opened 2018-05-20
Loosen bundle parsing logic

th0r:
Fixes https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/160


Koslun:
@th0r if "Some bundle chunks are missing", will it be reported somewhere or be immediately apparent what bundle is missing?

My possibly unfounded concern is that errors might get too silent and cause people to draw the wrong conclusions. Where I figure that if you have quite a lot of anonymous  bundles, one or two going missing from the graph could easily go unnoticed and thus ignored.

Where my assumption is that you'll be expecting the tool to perfectly match all the bundles and thus might not inspect that each bundle is there or even really look at your own list of bundle sizes, relying on this great tool instead.

valscion:
> will it be reported somewhere

Yes, this PR adds a warning console message to be printed when such chunks are analyzed. Before, the same warning used to fail the entire bundle analyze.

th0r:
@Koslun yes, it will report about it in the console: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/181/files#diff-22404c0ae8b5efcecea3b9b27cd95271R57

The thing is this is a very rare situation - absence of *some* chunks. In most cases you either don't have any chunks at all (I mean physical files to parse) or have them all and there shouldn't be any problems.

I wonder maybe we should introduce some kind of `strict: true` option?

PR #182, opened 2018-05-21
Update dependencies (including Webpack to v4)

th0r:
**Warning!** Drops Node v4 support!

### Do not merge until we decide to drop Node v4 support!


wanecek:
Any updates on this? :) I would personally really like seeing support for Webpack 4 :tada: Anything we can do to help?

wanecek:
Ah, excellent! Sorry for the misunderstanding :)

valscion:
No worries :blush:

valscion:
I think we can drop node4 support. We should also start using `bfj` instead of `bfj-node4` as the latter has now begun showing warnings after installations. See #198 and #196 

We can bump the `webpack-bundle-analyzer` major version when we do this.

Aladdin-ADD:
any progress?

th0r:
@Aladdin-ADD why do you need it so badly?

package.json
```diff
@@ -33,47 +33,48 @@
...
   "dependencies": {
-    "acorn": "^5.3.0",
-    "bfj-node4": "^5.2.0",
-    "chalk": "^2.3.0",
-    "commander": "^2.13.0",
-    "ejs": "^2.5.7",
-    "express": "^4.16.2",
-    "filesize": "^3.5.11",
+    "acorn": "^5.5.3",
+    "bfj-node4": "^5.3.1",
```
nktssh: if it will drop node 4 support, maybe, it worth to use `bfj` package instead? It seems that the current one is deprecated...
https://www.npmjs.com/package/bfj-node4

th0r:
All the dependencies will be updated in https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/206

PR #196, opened 2018-07-06
chore: bfj-node4 has been deprecated, use bfj

Aladdin-ADD:


valscion:
Thanks for letting us know `bfj-node4` is deprecated!

We will need to bump node support up from node 4 to do this upgrade. We will get to that at some point :relaxed:.

Aladdin-ADD:
Node 4 has been EOL for a while. Do you have any estimates on when to expect the new version to be published?

PR #200, opened 2018-07-18
documented using the plugin via the webpack cli

SKalt:
These docs are a nice-to-have.  I've found using the plugin via the webpack cli lets me analyze my webpack process without needing to create or modify a webpack config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,11 @@ module.exports = {
   ]
 }
 ```
+This is equivalent to the webpack 4+ CLI option 
+```sh
+webpack [...options] --plugin webpack-bundle-analyzer/lib/BundleAnalyzerPlugin
+```
+
 
 It will create an interactive treemap visualization of the contents of all your bundles.
 


valscion:
Hi! Thanks for opening the PR for discussion :relaxed:

Today I learned that webpack-cli has `--plugin` option 😅.

Importing/requiring anything other than the main entry point is not supported. Even though it works for you to use `webpack-bundle-analyzer/lib/BundleAnalyzerPlugin` as the path, it wasn't intended as a public API.

Is there a reason why you want to use the webpack CLI instead of the CLI usage we've documented for `webpack-bundle-analyzer` in README? https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/v2.13.1/README.md#usage-as-a-cli-utility

SKalt:
The only moderately compelling reason I've got for choosing to use  `--plugin` is avoiding running 
```sh
# analyze.sh
webpack [...entries, options] --profile --json > $some_path/stats.json \
  &&  webpack-bundle-analyzer $some_path/stats.json \
  && rm $some_path/stats.json;
```
which can run into cross-platform scripting/compatibility issues. 
My overall opinion: I'd love a more succinct option, but both the webpack and plugin clis work fine for me. 
Thanks for the quick reply!

valscion:
Interesting, I can see why you'd want to do that.

Have you considered adding the plugin to your webpack configuration instead, and protect it with some sort of conditional? For example:

```js
// webpack.config.js
const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');

module.exports = function (env) {
  // ...
  plugins: [
    (env.analyze && new BundleAnalyzerPlugin())
    // ...other plugins here
  ].filter(Boolean) // Filter out falsy values
}
```

...and compiling your code with something like this?

```bash
# analyze.sh
webpack --env.analyze [...options]
```

I think this is the correct CLI usage, if I read https://webpack.js.org/api/cli/#environment-options correctly.

SKalt:
That's probably the best way to to configure a reusable `npm run build-analyze` script for a project which already has a root-level webpack.config.js.

Still, for folks who are enjoying webpack 4 as a no-configuration bundler and/or want to do a one-off analysis `--plugin` is convenient:

```{sh}
# analyze just one bundle once without a webpack.config.js
npx webpack \
  ./src/entry/point.js \
  ./dist/temp.bundle.js \
  [...options] \
  --plugin webpack-bundle-analyzer/lib/BundleAnalyzerPlugin;

# existing wepback build pipeline set up with or without webpack.config.js:
npm run build -- --plugin webpack-bundle-analyzer/lib/BundleAnalyzerPlugin
```

SKalt:
Hi @valscion, I hit my use-case again: I needed to figure out why my latest bundle was bloated but didn't want to add/remove text to my webpack.config.js.  What would you think of using something like
```javascript
// webpack-bundle-analyzer/plugin.js
/*
This is intended for use as a webpack@^4 plugin:
`webpack [...options] --plugin webpack-bundle-analyzer/plugin` 
*/
module.exports = new (require('./lib/index.js').BundleAnalyzerPlugin)
```
to ensure using the plugin via the webpack cli (1) uses the supported api and (2) looks intuitive?

PR #205, opened 2018-08-14
Removed unnecessary folders from displaying in static analyzer file

varun-singh-1:
Currently, some projects with several nested folders cause the entire display to be taken up by folder names instead of actual files. For example, the path to the components might look like './build/docs/components/repo1/src/component1/component1.js', and all the extra folders would be displayed. Now, unless a folder has multiple children, it will be hidden (but the full path to each file can still be seen by hovering over the file).

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -124,11 +124,18 @@ function generateReport(bundleStats, opts) {
 
   if (!chartData) return;
 
+  const cleanChartData = chartData;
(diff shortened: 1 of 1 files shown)


th0r:
Thanks for the idea - implemented in https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/206.

PR #209, opened 2018-09-11
Remove unnecessary dependencies

### LINKED ISSUE #208: Missing `preact` dependency?
#### Issue description
When installing v3.0.1, the following warning appears:
`warning "webpack-bundle-analyzer > mobx-preact@3.0.0" has unmet peer dependency "preact@>=8".`

#### Technical info
* Webpack Bundle Analyzer version: 3.0.1
* Webpack version: 4.18.0
* Node.js version: 8.12.0
* npm/yarn version: yarn 1.9.4
* OS: fedora 28

realityking:
Love the v3 release, the new output is much cleaner. Great work!

I noticed v3 pulls in a few new dependencies that aren't necessary. This PR fixes that.

- Move those dependencies that are only used in the client to dev-dependencies. Meaning `@babel/polyfill`, `mobx` and `mobx-preact`. Also fixes #208.
- Increase the target node version in the babel config to 6.14.4. There are some plug-ins that fall off when targeting >= 6.5 so this helps load less polyfills.
- Remove transform-runtime and thus the dependency on `@babel/runtime` for the plugin itself (it stays around for the client). It only loaded four functions from the runtime (all related to ESM support) but brings in heavy dependencies.
- Last but not least, adds the semver operator `^` for the dependency on `lodash`. It was the only one that didn't have it.


realityking:
@th0r Ups, that wasn't intentional. Fixed.

th0r:
Thanks a lot!

th0r:
Released in `v3.0.2`

PR #212, opened 2018-09-28
Use polka instead of express to serve the viewer

realityking:
While the code is less elegant, this is a pretty big size reduction for anyone not using express as you can see in the number of dependencies that become dev dependencies. On the other hand, it's a relatively small size increase for those projects that do use polka.

Express is still the most popular web framework but I wouldn't assume that everyone is using it since there are so many great alternatives. Since bundle analyser has simple needs, something tiny like polka should do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,11 +37,12 @@
     "acorn-walk": "^8.0.0",
     "chalk": "^4.1.0",
     "commander": "^6.2.0",
-    "express": "^4.17.1",
     "filesize": "^6.1.0",
     "gzip-size": "^6.0.0",
     "lodash": "^4.17.20",
     "opener": "^1.5.2",
+    "polka": "^0.5.2",
+    "sirv": "^1.0.7",
     "ws": "^7.3.1"
   },
   "devDependencies": {
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -4,7 +4,8 @@ const http = require('http');
 
 const WebSocket = require('ws');
 const _ = require('lodash');
-const express = require('express');
+const polka = require('polka');
+const sirv = require('sirv');
 const {bold} = require('chalk');
 
 const Logger = require('./Logger');
@@ -48,8 +49,8 @@ async function startServer(bundleStats, opts) {
 
   if (!chartData) return;
 
-  const app = express();
-  app.use(express.static(`${projectRoot}/public`));
+  const app = polka();
+  app.use(sirv(`${projectRoot}/public`));
 
   app.get('/', (req, res) => {
     res.writeHead(200, {'Content-Type': 'text/html'});
@@ -63,7 +64,7 @@ async function startServer(bundleStats, opts) {
     return res.end(html);
   });
 
-  const server = http.createServer(app);
+  const server = http.createServer(app.handler);
 
   await new Promise(resolve => {
     server.listen(port, host, () => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,11 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@arr/every": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@arr/every/-/every-1.0.1.tgz",
+      "integrity": "sha512-UQFQ6SgyJ6LX42W8rHCs8KVc0JS0tzVL9ct4XYedJukskYVWTo49tNiMEK9C2HTyarbNiT/RVIRSY82vH+6sTg=="
+    },
     "@babel/code-frame": {
       "version": "7.10.4",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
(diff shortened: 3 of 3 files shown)


src/viewer.js
```diff
@@ -39,24 +45,26 @@ async function startServer(bundleStats, opts) {
...
+  app.get('/', (req, res) => {
+    const data = {
       mode: 'server',
       get chartData() { return JSON.stringify(chartData) },
       defaultSizes: JSON.stringify(defaultSizes)
+    };
+
+    ejs.renderFile('views.ejs', data, ejsOptions, (err, str) => {
+      if (err) {
+        throw err;
```
th0r: Shouldn't `err` be passed to the `next` route handler?

package.json
```diff
@@ -38,12 +38,13 @@
     "chalk": "^2.4.1",
     "commander": "^2.18.0",
     "ejs": "^2.6.1",
-    "express": "^4.16.3",
     "filesize": "^3.6.1",
     "gzip-size": "^5.0.0",
     "lodash": "^4.17.10",
     "mkdirp": "^0.5.1",
     "opener": "^1.5.1",
+    "polka": "^0.5.1",
```
th0r: How does it reduce the total size of dependencies? I would like to see the numbers.

realityking:
Fair question on the size. Using bundle phobia:

[webpack-bundle-analyzer](https://packagephobia.now.sh/result?p=webpack-bundle-analyzer) is currently at 6.05MB, [express](https://packagephobia.now.sh/result?p=express) is 1.56 MB so a quarter of that.

[polka](https://packagephobia.now.sh/result?p=polka) is just 53kB and [serve-static](https://packagephobia.now.sh/result?p=serve-static) is 318kB.

So this PR would shave off a bit more than 1MB of the install size.

src/viewer.js
realityking: This isn't a middleware but a handler for `GET` and thus doesn't have a `next` handler. I could output an error to the browser but I figured this be enough for the use case of bundle analyzer.

th0r:
For me `node_modules` sizes using `npm ci --production` are `10mb` with `express` and `8.3mb` with `polka`. Although it's a quite significant difference I would better wait for the polka to reach v1 and stabilize it's API before switching to it.

realityking:
@th0r I don't think polka will receive a 1.0 anytime soon. It's more likely one of those projects that (annoyingly) stays < 1.0 forever. 

Since polka has a very small API and bundle analyzer makes only 3 calls to it, I think migrating to a new API would be easy.

We could pin to a fixed version if that helps. What do you think?

lukeed:
Hi! FWIW `polka@next` (and all others publicly available with the `next` tag) are completely stable and have been for quite some time. They're already powering multiple high-traffic production applications. 
The only reason `1.0` hasn't been released yet is because I'm finalizing **new** module packages and adding a documentation site. The new packages would focus on new alternatives for things like cookie handling, compression, etc. What's publicly available now is finalized 👍

package.json
```diff
@@ -39,12 +39,13 @@
...
     "commander": "^2.18.0",
     "ejs": "^2.6.1",
-    "express": "^4.16.3",
     "filesize": "^3.6.1",
     "gzip-size": "^5.0.0",
     "lodash": "^4.17.15",
     "mkdirp": "^0.5.1",
     "opener": "^1.5.1",
+    "polka": "^0.5.2",
+    "serve-static": "^1.14.1",
```
lukeed, line 48: I would suggest [`sirv`](https://github.com/lukeed/sirv) here instead for further size (and performance) improvements. Unfortunately `serve-static` is a significant portion of Express' size. You can find an example with Polka [here](https://github.com/lukeed/polka/blob/master/examples/with-sirv/index.js)

src/viewer.js
lukeed: With Polka@next, all route handlers follow (req, res, err) signature, so you're no longer limited here.

Here's an [alternative approach](https://github.com/lukeed/polka/issues/73#issuecomment-428315454) using the current `polka@0.5` version

package.json
lukeed: https://bundlephobia.com/result?p=polka@0.5.2
VS
https://bundlephobia.com/result?p=express@4.17.1

If/when serve-static is replaced:

https://bundlephobia.com/result?p=sirv@1.0.1
VS
https://bundlephobia.com/result?p=serve-static@1.14.1

realityking:
I've rebased this PR to be mergeable again. And also adopted `sirv` as suggested by @lukeed.

I've did stick to `polka@^0.5.2` instead of using the preview releases.

@lukeed It'd be great to a stable polka 1.0 :)

realityking:
Comparison of `node_modules` sizes:

* This branch: 39 packages weighing 3,9 MB
* Master: 80 packages weighing 5,4 MB

realityking:
Updated this in light of #397 being merged.

New comparison of `node_modules`:

- This branch: 34 packages weighing 4,4 MB
- master: 65 packages weighing 5,9 MB

Still ~half the packages and 20% less bytes.

(I have no idea why the size went up compared to the last time I ran the stats)

valscion:
Thank you for the contributions! In the end, we went with #398 :relaxed:

realityking:
All good. Thank you :)

PR #215, opened 2018-10-16
Disable viewer websocket connection if mode is static

### LINKED ISSUE #195: Websocket Connection blocked when serving over HTTPS
The opened WebSocket connection is always `ws:`. If the file is served via HTTPS, this should be `wss:` or else the browser will block the connection.

Message from Chrome:

    bundle-report.html:formatted:967 Mixed Content: The page at 'https://foo.bar/bundle-report.html'
    was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://foo.
    bar/'. This request has been blocked; this endpoint must be available over WSS.

The culprit is here, probably:
https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/2c0555536e215c8096ed6f7d10b1e5783251f76a/client/viewer.jsx#L11

sebastianhaeni:
Resolves #195

### DIFF
--- a/client/viewer.jsx
+++ b/client/viewer.jsx
@@ -9,7 +9,9 @@ import styles from './viewer.css';
 // Initializing WebSocket for live treemap updates
 let ws;
 try {
-  ws = new WebSocket(`ws://${location.host}`);
+  if (window.enableWebSocket) {
+    ws = new WebSocket(`ws://${location.host}`);
+  }
 } catch (err) {
   console.warn(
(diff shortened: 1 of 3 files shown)


valscion:
Looks good to me. What do you think, @th0r?

th0r:
Looks good. Thanks @sebastianhaeni!
@valscion could you publish a new version with this change because I don't have time right now?

valscion:
>  @valscion could you publish a new version with this change because I don't have time right now?

Will do :+1:

valscion:
Published v3.0.3 with this change :tada:

PR #217, opened 2018-10-21
Proposal: allow bundle content retrieval to be customized in `parseBundle`

andrewbranch:
I'm looking to use some parts of this project in a scenario where it's easiest to retrieve bundle files from a CDN rather than from the file system itself. I could download from the CDN to a temporary location, but by abstracting the `fs` calls to a “bundle host” object, I can feed the function a string from memory rather than relying on the file system itself. It would be even nicer if this function were async, and I could just tell the bundleHost to `fetch()` from the CDN, but I figured adding that complexity wouldn’t be warranted here.

At the moment, I only need `parseBundle`, and so only made changes there, but this approach could be used throughout the project, and I'd be open to making those changes if that would make this PR seem more appealing.

Thoughts? Thanks for this project and for the consideration!

### DIFF
--- a/src/parseUtils.js
+++ b/src/parseUtils.js
@@ -7,8 +7,12 @@ module.exports = {
   parseBundle
 };
 
-function parseBundle(bundlePath) {
-  const content = fs.readFileSync(bundlePath, 'utf8');
+const defaultBundleHost = {
+  getFileContent: path => fs.readFileSync(path, 'utf8')
+};
+
+function parseBundle(bundlePath, bundleHost = defaultBundleHost) {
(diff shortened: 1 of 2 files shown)


andrewbranch:
Glancing through the issues, this seems tangentially related to #32, since my goal in making this change is to use some of the analyzer code without the UI.

andrewbranch:
Cleaning up some loose ends this morning, going to close this one out. I no longer need this change and it seems like no one else does either. If there's interest in the future feel free to resurrect this code and use it. Cheers! 👋🏼

PR #221, opened 2018-11-07
update Webpack profile command to set reporter as json

billymoon:


valscion:
Thanks for the PR! Do you know which version is the first one supporting the `--reporter json`? I'm asking because I'm not sure if we should document both versions in case we've got plenty of users on old webpack versions.

billymoon:
@valscion well, I just checked, and turns out I am using the wrong webpack cli, I was using `webpack-command` but should have been using `webpack-cli` which does have the `--json` flag.

valscion:
Ok thanks for checking! :+1:

PR #223, opened 2018-11-15
replace <str>.length with Buffer.byteLength(<str>)

hulkish:
Some chars represent a byte length greater than 1. This means string length could report incorrect `parsedSize`, for bundles - if any special chars are used. More on this [here](https://sankhs.com/2016/03/17/content-length-http-headers/).


EugeneHlushko:
Looks legit, do you think you could add an auto test to cover this case?

hulkish:
@valscion can you point me in the right direction for this? Im trying to get an idea of how i am intended to generate the fixture for this test?

valscion:
The best bet might be to create a small reproduction repository with plain webpack and generate a stats.json object and output bundle from there. Supply that output to the basis of your new test and test that the expected sizes match what you want.

hulkish:
@valscion that makes sense, thank you. 

I guess I was looking for a script that creates one for me, or something lol

hulkish:
@valscion @EugeneHlushko will this do it?

th0r:
@hulkish thanks, merged!

PR #224, opened 2018-11-24
add ts typings

whitecolor:
There where well-documented typings on `@types/..`,  I corrected them a little, I believe it typings always should go with a project repo and package.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,88 @@
+// Definitions by: Michael Strobel <https://github.com/kryops>
+
+import * as webpack from "webpack";
+
+export type ExcludeAssetsPatternFn = (assetName: string) => boolean;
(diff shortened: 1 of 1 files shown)


whitecolor:
image

valscion:
Thanks, but maintaining typescript types in this package when we don't consume them ourselves would incur a maintenance cost that we would prefer not to have. The `@types/` project is suitable for this case.

whitecolor:
But you don't need to maintain them, besides types are very simple there will be no problems with them, it is much more convenient for users.

valscion:
See https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/83 for previous discussion.

@th0r might want to also answer if we want TS typings here or not

PR #226, opened 2018-11-25
defaultPinned property

nhtb:
Sidebar closes after every refresh even if it's pinned and it disappears after 3 seconds.
I prefer to use sidebar stats during bundle-optimization and it makes harder. So, I've added defaultpinned property to prevent this behaviour


nhtb:
I will recommit properly

maurocolella:
This would really really be useful.

PR #231, opened 2018-12-05
Update README.md (#191)

### LINKED ISSUE #191: Document webpack-dev-server interplay only displaying stat sizes
The closed issue #147 has some valuable comments about `webpack-bundle-analyzer` only showing stat sizes when bundle output does not go to the filesystem.

It would be valuable to document this caveat in the README.md.

## This contribution opportunity has been seized by @fanich37

**Please avoid creating competing pull requests**

---

Here's some background from the linked issue:

@jeron-diovis in https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/147#issuecomment-380737178

> Just ran into this issue using analyzer with `webpack-dev-server`. No angular or things like this. 
> 
> The problem is that analyzer uses physical files to calc parsed/gzipped sizes – while `webpack-dev-server` keeps everything in memory.
> 
> Second problem is that being used as plugin, analyzer shows no errors, while CLI tools says `Couldn't parse bundle asset`.
> 
> As I understand, there is nothing analyzer can do with `webpack-dev-server`, but maybe it's worth to explicitly mention in docs that you should have files on disc to get parsed/gzipped info?

@AgentME in https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/147#issuecomment-397444503

> I was using webpack with gulp and `compiler.outputFileSystem = new MemoryFS();` so that webpack wouldn't write the bundles itself, so I ran into the same issue as @jeron-diovis. The BundleAnalyzerPlugin didn't report any errors at all in the console, just silently produced stat-sized-only reports. I had to use `webpack-bundle-analyzer` on the command line on the profile json in order to see any kind of error message, which I only thought to do because of this thread.

fanich37:
Update information that plugin provides `parsed` and `gzip` size values only if files actually exist in the file system.

Closes #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,8 @@ ng build --stats-json --build-optimizer=false --vendor-chunk=true
 ```
 The resulting stats.json can be found at `dist/stats.json`
 
+*Please be aware, that this is not a reasonable build for production use, as it increases the build output size considerably.*
+
 ### I don't see `gzip` or `parsed` sizes, it only shows `stat` size
 
 It happens when `webpack-bundle-analyzer` analyzes files that don't actually exist in your file system, for example when you work with `webpack-dev-server` that keeps all the files in RAM. If you use `webpack-bundle-analyzer` as a plugin you won't get any errors, however if you run it via CLI you get the error message in terminal:
@@ -173,7 +175,6 @@ Analyzer will use module sizes from stats file.
 ```
 To get more information about it you can read [issue #147](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/147).
 
-*Please be aware, that this is not a reasonable build for production use, as it increases the build output size considerably.*
 
 <h2 align="center">Maintainers</h2>
 
--- a/README.md
+++ b/README.md
@@ -166,6 +166,15 @@ ng build --stats-json --build-optimizer=false --vendor-chunk=true
 ```
 The resulting stats.json can be found at `dist/stats.json`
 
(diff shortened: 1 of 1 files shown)


valscion:
Well, I think it's closer to "troubleshooting" than usage ¯\\\_(ツ)_/¯. I don't have strong opinions on this, but I'd like us to avoid adding too much documentation right in the start for edge cases

th0r:
Agree with @valscion here - "troubleshooting" section seems to be the right place.

fanich37:
Thanks for opinions @thor and @valscion . I'll redo it!

README.md
```diff
@@ -164,6 +164,15 @@ ng build --stats-json --build-optimizer=false --vendor-chunk=true
...
+### I don't see `gzip` or `parsed` sizes, it only shows `stat` size
+
+It happens when `webpack-bundle-analyzer` analyzes files that don't actually exist in your file system, for example when you work with `webpack-dev-server` that keeps all the files in RAM. If you use `webpack-bundle-analyzer` as a plugin you won't get any errors, however if you run it via CLI you get the error message in terminal:
+```
+Couldn't parse bundle asset "your_bundle_name.bundle.js".
+Analyzer will use module sizes from stats file.
+```
+To get more information about it you can read [issue #147](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/147).
+
 *Please be aware, that this is not a reasonable build for production use, as it increases the build output size considerably.*
```
th0r: I think this line refers to the previous section.

th0r:
@fanich37 thanks a lot!

PR #242, opened 2019-01-19
feat: add brotli size support

### LINKED ISSUE #113: Add brotli size support

manniL:
Hey :wave:

As native Brotli support landed in the latest minor Node version, I thought it wouldn't hurt to file a PR that leverages the support to calculate Brotli sizes (similar to #114 but with the native implementation used). I approached the problem similar to the GZIP implementation (using the Brotli version of [gzip-size](https://github.com/sindresorhus/gzip-size#readme) called [brotli-fsize](https://github.com/Developmint/brotli-fsize/).

## Problems

As the **native implementation** is only present in Node 11.7.0 and upwards (might be backported to the LTS at on day but even this doesn't support Node <= 10.X) and the used package explicitly asks for the version, I'm not sure how to deal with lower Node versions here.

I could remove the strict version requirements and check the node version in this package instead but this seems more like a hacky solution.

I'm waiting for your feedback/ideas :relaxed: 

Resolves #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ Directory containing all generated bundles.
   -p, --port <n>              Port that will be used in `server` mode to start HTTP server. (default: 8888)
   -r, --report <file>         Path to bundle report file that will be generated in `static` mode. (default: report.html)
   -s, --default-sizes <type>  Module sizes to show in treemap by default.
-                              Possible values: stat, parsed, gzip (default: parsed)
+                              Possible values: stat, parsed, gzip, brotli (default: parsed)
   -O, --no-open               Don't open report in default browser automatically.
   -e, --exclude <regexp>      Assets that should be excluded from the report.
                               Can be specified multiple times.
@@ -147,6 +147,11 @@ as Uglify, then this value will reflect the minified size of your code.
 ### `gzip`
 
 This is the size of running the parsed bundles/modules through gzip compression.
+
+
+### `brotli`
+
+This is the size of running the parsed bundles/modules through brotli compression.
 
 <h2 align="center">Troubleshooting</h2>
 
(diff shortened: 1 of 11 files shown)


th0r:
Hey! Thanks for the PR but I see a few problems here.

First one is, as you mentioned, this PR requires Node 11.7.0 and upwards but we must support Node.js >= 6.14.4.

The second one is I'm not sure a lot of people are interested in Brotli sizes, but it's calculation takes quite a lot of analyze time. As @valscion [mentioned](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/113#issuecomment-327717813) in #113 the best solution would be to have an option that would enable calculation of Brotli sizes.

th0r:
> I could lower the package requirements

Do you mean requirements of the `brotli-fsize`? Is it yours lib?

> either to enable the feature or to disable it

I think it should be disabled by default. We can reconsider it for the next major version of the analyzer.

manniL:
> Do you mean requirements of the `brotli-fsize`? Is it yours lib?

@th0r Yup, it's my lib 😄 (https://www.npmjs.com/package/brotli-fsize)

> I think it should be disabled by default. We can reconsider it for the next major version of the analyzer.

Fair enough! Will look into adding the flag option 👍

ElForastero:
Any updates?

278kunal:
Hi, team when are we merging this?

valscion:
See https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/113#issuecomment-515996212

manniL:
Closing due to inactivity

PR #247, opened 2019-02-02
Make webpack's done hook wait until analyzer writes report / stat file.

### LINKED ISSUE #232: `setImmediate` causes analyzer to only start after webpack's `done` callback is invoked
### Issue description
This issue stems from [parallel-webpack#85](https://github.com/trivago/parallel-webpack/issues/85) and is related to #152, and a PR in [parallel-webpack#89](https://github.com/trivago/parallel-webpack/pull/89) in an attempt to 'hack around' the faulty behaviour.

`webpack-bundle-analyzer` is enqueueing the main analyze operation with [`setImmediate`](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/src/BundleAnalyzerPlugin.js#L58-L60), which affects this plugin to execute only after a webpack build (which is executed with the Node.js API) has invoked the `done` callback.

I suspect that when webpack is running, there are a lot of I/O operations being spawned by webpack (like reading and writing), therefore the actions that are defined and called using `setImmediate` are deferred after the webpack build has finished. Since there is no way to 'listen' when this plugin has successfully completed, **this causes issues with tools that exits as soon as the `done` callback is fired (for example: tooling that spawns multiple webpack builds in parallel and hooks into the `done` callback to determine if they are done):**

```js
const webpack = require('webpack')
const config = require('./webpack.config')

const compiler = webpack(config)

compiler.run(/* done callback */ (err, stats) => {
    if (err) {
        console.log(err)
        process.exit(1)
    }
    console.log('build succeeded')
    process.exit(0)
    // now the analyzer will not be able to emit anything here
})
```

### Reproducible repository
I've created a minimal repository that reproduces this issue, and a detailed explanation here: https://github.com/liangchunn/webpack-bundle-analyzer-set-immediate-bug

Please read the README as it contains important things to watch out in order to understand how it actually works.

### Possible fixes
- Don't enqueue operations with `setImmediate`, make it execute whenever the line is reached. This would change the behaviour of how people use it, and might echo logs between builds (Reference #21)
- Have an option like `{ deferAnalyzer: boolean = true}` (true by default), and explicitly opt-out enqueuing operations with `setImmediate` by setting `{ deferAnalyzer: false }`, which won't break everyone's build or change the way people are already using this plugin

**In any case, I am open for a PR addressing this issue. Just let me know which route to go (or if there are any other possibilities to fix this issue).**

#### Technical info
* Webpack Bundle Analyzer version: 3.0.3
* Webpack version: 4.27.1
* Node.js version: 8.11.3
* yarn version: 1.12.3
* OS: macOS 10.14.1

#### Debug info
**plugin**

```js
{
    analyzerMode: 'static',
    reportFilename: './report.html',
    openAnalyzer: false
}
```

- clean-webpack-plugin

mareolan:
Fixes #232 by prolonging webpack's "done" hook until analyzer finishes writing report / stat files.

### DIFF
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -34,6 +34,7 @@ class BundleAnalyzerPlugin {
     this.compiler = compiler;
 
     const done = (stats, callback) => {
+      callback = callback || (() => {});
       stats = stats.toJson(this.opts.statsOptions);
 
       const actions = [];
@@ -57,14 +58,14 @@ class BundleAnalyzerPlugin {
         // Making analyzer logs to be after all webpack logs in the console
         setImmediate(async () => {
           try {
-            for (let i = 0; i < actions.length; ++i) await actions[i]();
-            if (typeof callback === 'function') callback();
+            await Promise.all(actions.map(action => action()));
+            callback();
           } catch (e) {
-            if (typeof callback === 'function') callback(e);
+            callback(e);
           }
         });
       } else {
-        if (typeof callback === 'function') callback();
+        callback();
       }
     };
 
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -33,7 +33,7 @@ class BundleAnalyzerPlugin {
   apply(compiler) {
     this.compiler = compiler;
 
-    const done = stats => {
+    const done = (stats, callback) => {
       stats = stats.toJson(this.opts.statsOptions);
 
       const actions = [];
@@ -55,14 +55,19 @@ class BundleAnalyzerPlugin {
 
       if (actions.length) {
         // Making analyzer logs to be after all webpack logs in the console
-        setImmediate(() => {
-          actions.forEach(action => action());
+        setImmediate(async () => {
+          try {
+            for (let i = 0, len = actions.length; i < len; ++i) await actions[i]();
+            if (typeof callback === "function") callback();
+          } catch (e) {
+            if (typeof callback === "function") callback(e);
+          }
         });
       }
     };
 
     if (compiler.hooks) {
-      compiler.hooks.done.tap('webpack-bundle-analyzer', done);
+      compiler.hooks.done.tapAsync('webpack-bundle-analyzer', done);
     } else {
       compiler.plugin('done', done);
     }
@@ -108,8 +113,8 @@ class BundleAnalyzerPlugin {
     }
   }
 
-  generateStaticReport(stats) {
-    viewer.generateReport(stats, {
+  async generateStaticReport(stats) {
+    await viewer.generateReport(stats, {
       openBrowser: this.opts.openAnalyzer,
       reportFilename: path.resolve(this.compiler.outputPath, this.opts.reportFilename),
       bundleDir: this.getBundleDirFromCompiler(),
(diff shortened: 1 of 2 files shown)


liangchunn:
Nice work! This fixes the issue in my reproducible repo mentioned in #232.

src/BundleAnalyzerPlugin.js
```diff
@@ -55,14 +55,19 @@ class BundleAnalyzerPlugin {
 
       if (actions.length) {
         // Making analyzer logs to be after all webpack logs in the console
-        setImmediate(() => {
-          actions.forEach(action => action());
+        setImmediate(async () => {
+          try {
+            for (let i = 0, len = actions.length; i < len; ++i) await actions[i]();
```
liangchunn: I don't think we need to pre-optimise the loop here. AFAIK V8's optimising compiler does this for us automatically.

mareolan: You mean write it as `for (let i = 0; i < actions.length; ...` instead?

liangchunn: Yes, IMO `for(let i = 0; i < actions.length; i++)` more readable than `for(let i = 0, len = actions.length; i < len; ++i )`, plus it wouldn't make a huge difference in terms of perf.

mareolan: Sure, no problem.

valscion:
This is looking good to me, thanks! Have you checked which version of `webpack` has the `tapAsync` API available, i.e. is it available in webpack 3 or only in 4 and up?

mareolan:
`tapAsync` is available only for webpack >= 4.0.0. I tried the repro also with latest 3.12.0 and since there's no `compiler.hooks`, it uses the alternative code flow and webpack won't wait for bundle analyzer. So basically, this fix fixes it only for webpack >= 4.0.0.

src/BundleAnalyzerPlugin.js
```diff
@@ -55,14 +55,21 @@ class BundleAnalyzerPlugin {
 
       if (actions.length) {
         // Making analyzer logs to be after all webpack logs in the console
-        setImmediate(() => {
-          actions.forEach(action => action());
+        setImmediate(async () => {
+          try {
+            for (let i = 0; i < actions.length; ++i) await actions[i]();
+            if (typeof callback === 'function') callback();
```
th0r: Let's replace these `if` conditions with `callback = callback || () => {};` at the top of the `done` function.
1/3

src/BundleAnalyzerPlugin.js
```diff
@@ -55,14 +55,21 @@ class BundleAnalyzerPlugin {
 
       if (actions.length) {
         // Making analyzer logs to be after all webpack logs in the console
-        setImmediate(() => {
-          actions.forEach(action => action());
+        setImmediate(async () => {
+          try {
+            for (let i = 0; i < actions.length; ++i) await actions[i]();
```
th0r: Those actions are not related to each other and can be run concurrently. So it can be replaced with `await Promise.all(actions.map(action => action())`.

src/BundleAnalyzerPlugin.js
```diff
@@ -55,14 +55,21 @@ class BundleAnalyzerPlugin {
...
+            if (typeof callback === 'function') callback();
+          } catch (e) {
+            if (typeof callback === 'function') callback(e);
```
th0r: 2/3

src/BundleAnalyzerPlugin.js
```diff
@@ -55,14 +55,21 @@ class BundleAnalyzerPlugin {
...
+          }
         });
+      } else {
+        if (typeof callback === 'function') callback();
```
th0r: 3/3

valscion:
This is now released in v3.0.4. Thank you for your contributions!

PR #248, opened 2019-02-04
Update acorn to version 6

### LINKED ISSUE #245: Potentially update `acorn` to match latest `webpack`
#### Issue description

I recently encountered webpack/webpack#8656 during an upgrade where, due to an [NPM bug](https://github.com/npm/cli/pull/147), the wrong version of `acorn` "wins" and breaks `webpack`. Running `npm ls acorn` shows the "winner" of the hoisting being `webpack-bundle-analyzer` (which I have in devDeps to make analyzing and testing easier). 

The `npm update acorn --depth 20 && npm dedupe` solution does not work in this case, however manually installing via `npm install --save-dev acorn@latest` worked.

Since `webpack-bundle-analyzer` is used rarely (only when manually invoked) I may be moving forward by removing it from my package.json, which is probably best practice anyways. I simply wanted to file the issue to add more info/color if anyone else encounters this when they upgrade their webpack.

#### Technical info
* Webpack Bundle Analyzer version: 3.0.3
* Webpack version: 4.29.0
* Node.js version: 10.14.1
* npm/yarn version: npm 6.7.0
* OS: macOS 10.14.3

#### Debug info

_Actual functionality of the tool is unaffected_

realityking:
Fixes #245

Supporting both acorn 5 and acorn 6 is not feasible because `acorn-walk` got split into its own module.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1107,9 +1107,9 @@
       }
     },
     "acorn": {
-      "version": "5.7.3",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.7.3.tgz",
-      "integrity": "sha512-T/zvzYRfbVojPWahDsE5evJdHb3oJoQfFbsrKM7w5Zcs++Tr257tia3BmMP8XYVjp1S9RZXQMh7gao96BlqZOw=="
+      "version": "6.0.7",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-6.0.7.tgz",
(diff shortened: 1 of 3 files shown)


th0r:
Won't it break setups with webpack 3?

domoratskii:
Hi! Are you going to merge ?

realityking:
@th0r  I don't see why it would `webpack-bundle-analyzer` parses the JS file itself and does not rely on the AST generated by webpack. Bundle analyzer will support more language features than webpack itself but that's a much better situation than the opposite.

th0r:
@realityking yep, you're right. Merged. Thanks!

Issue #229, opened 2018-11-30
Support for JavaScript modules


JonathanWbn:
Would it make sense to add support for files with the `.mjs` extension?

If so, I'd be happy to take a stab at this.

valscion:
Fixed by #252

PR #252, opened 2019-03-04
support for mjs

jlopezxs:
#229


valscion:
Please remove `yarn.lock`. You will also need to sign the CLA

src/analyzer.js
```diff
@@ -10,6 +10,7 @@ const {parseBundle} = require('./parseUtils');
 const {createAssetsFilter} = require('./utils');
 
 const FILENAME_QUERY_REGEXP = /\?.*$/;
+const FILENAME_EXTENSIONS = /.(js|mjs)$/i;
```
valscion: This would match `fooXjs`, too: You will need to escape the `.` as it matches any character:

```suggestion
const FILENAME_EXTENSIONS = /\.(js|mjs)$/i;
```

valscion:
Thanks! Looks nice. Do you mind adding a new changelog entry right under the

```
## UNRELEASED

<!-- Add changelog entries for new changes under this section -->
```

`<!--` comment?

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 ## UNRELEASED
 
 <!-- Add changelog entries for new changes under this section -->
+<!--
+* **Improvement**
+   * Add support JS modules ([#252](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/252), [@jlopezxs](https://github.com/jlopezxs)))
+-->
```
valscion: Ah, you don't need to add the new changelog entry in a comment :relaxed:. Only the "Add changelog entries" line needs to be a comment, new entries can be listed as is.

Also, what do you think if the wording was changed to "Add support for .mjs output files"? I fear that "support for JS modules" would cause people to think that `webpack-bundle-analyzer` didn't support JS modules as sources before. Because this PR is about supporting `.mjs` output files, right?

valscion:
Looks good to me :+1:. Thanks for contributing!

valscion:
This is now released in webpack-bundle-analyzer v3.2.0. Thank you for your contributions!

PR #262, opened 2019-04-11
Properly escape embedded JS/JSON

th0r:
Don't let `</script>` in `chartData` or in internal assets break the page.

Vulnerability details: https://blog.uploadcare.com/vulnerability-in-html-design-the-script-tag-33d24642359e


valscion:
This seems to have caused a regression, see #263

Issue #263, opened 2019-04-11
Invalid or unexpected token


paradoxxxzero:
# Note from maintainers: Yes, we know about this regression. Avoid posting +1 comments please

Please use version v3.3.0 in the meantime.

---

Something seems wrong with the last release (3.3.1) :

![image](https://user-images.githubusercontent.com/271144/55961374-0f571000-5c6f-11e9-9b5b-85a57d8abacc.png)

valscion:
Have you checked that it was working with v3.3.0?

paradoxxxzero:
Yes it works

valscion:
Ok @th0r it's likely that https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/262 might have caused an issue. There's a strange `\u003c` line in the code that should be valid JS

valscion:
I'm able to reproduce this on our own repository when building a static report.

![Developer_Tools_-_Webpack_Bundle_Analyzer_-_file____Users_vesa_code_v_venuu_tmp_performance-reports_1554990096-report_html](https://user-images.githubusercontent.com/482561/55962088-e8054080-5c78-11e9-9880-9d098530f9c9.png)

valscion:
Server build has no issues

crazyx13th:
![grafik](https://user-images.githubusercontent.com/8912561/55962948-28ad8b80-5c72-11e9-9be4-4db750f139f0.png)
same here, firefox and chrome... thx

valscion:
Yup we're looking into this. This is a known regression

th0r:
Fixed in v3.3.2

PR #264, opened 2019-04-11
Fix regression with escaping internal assets

th0r:
Fixes https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/263


valscion:
Yup it fixed the regression for me in our project. Thanks!

PR #267, opened 2019-04-15
Use native Object

Connormiha:
Just used native Object's methods


valscion:
Thanks — the existing code is fine.

PR #271, opened 2019-04-23
output the report.html to memoryfs when it is from webpack-dev-server

### LINKED ISSUE #270: Feature Request: output the report.html to memoryfs when it is from webpack-dev-server
#### Issue description
Actually, this lib should not support "server" scenario because webpack-dev-server can do this for us. What you need is to output the report.html to the MemoryFS. It also solves #12

wood1986:
fix #270 and #12


th0r:
Thanks for the contribution, but I don't want to introduce breaking change and force people to install `webpack-dev-server` if they don't want to generate report files in filesystem.

Issue #274, opened 2019-04-25
Allow disabling of bundleDir from plugin option


amoshydra:
#### Issue description
Allow us to disable/overwrite `bundleDir` in the plugin option

Currently `bundleDir` can only be defined in via the command line. In plugin `bundleDir` is decided in the following logic

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/1c8aba35c62850a2e30a978471efba8f3925b643/src/BundleAnalyzerPlugin.js#L130-L132

However, `compiler.outputFileSystem` might not always be `MemoryFileSystem`. 

##### Use case:
The framework `Nuxt 2.5.0` [replace](https://github.com/nuxt/nuxt.js/blob/fbabeb8fce4630c193416e178f10f75d841083b5/packages/webpack/src/utils/async-mfs.js) `MemoryFileSystem` with their own implementation of `AsyncMFS`, which extend `MemoryFileSystem`, during development.

##### Current problem
This cause `compiler.outputFileSystem.constructor.name` to be `"AsyncMFS"`, thus causing`getBundleDirFromCompiler` to return `this.compiler.outputPath`. 

##### Expected behavior
The expectation result here is `null`. Since nuxt is using `webpack-dev-server` during development and no real file is being created.

##### My thoughts
I am not sure if it is possible to detect if real file is created in this case. It might not be reasonable for `webpack-bundle-analyzer` to keep track of a list of `outputFileSystem.constructor.name` that should make `bundleDir` to `null`.

Thus I am proposing `webpack-bundle-analyzer` to allow us to overwrite `bundleDir` to `null` in the plugin option

#### Technical info
* Webpack Bundle Analyzer version: v3.3.2
* Webpack version: v4.30.0
* Node.js version: v8.11.3
* npm/yarn version: v1.12.1
* OS: Windows 10, 1803 (OS Build 17134.706)

#### Debug info
How do you use this module? As CLI utility or as plugin?

> As plugin

If CLI, what command was used? (e.g. `webpack-bundle-analyzer -O path/to/stats.json`)

If plugin, what options were provided? (e.g. `new BundleAnalyzerPlugin({ analyzerMode: 'disabled', generateStatsFile: true })`)

```
new webpackBundleAnalyzer.BundleAnalyzerPlugin({
  openAnalyzer: false,
}),
```
What other Webpack plugins were used?

> This plugin is used with https://github.com/nuxt/nuxt.js `v2.6.2`

[...]

PR #275, opened 2019-04-25
Add AsyncMFS for getBundleDirFromCompiler

amoshydra:
### Details
- Detect `AsyncMFS` to support dev-server of Nuxt 2.5 and above

### Related issues
- #274

PR #276, opened 2019-04-25
Allow overwriting of `bundleDir` via `bundleDirOption` from plugin option.

amoshydra:
### Details
- Allow overwriting of `bundleDir` via `bundleDirOption` from plugin option.
- Updated documentation for `bundleDirOption`
- Added 3 tests for `bundleDirOption`

```
  Plugin
    options
      bundleDirOption
        √ should report parsedSize and gzipSize when bundleDirOption is not provided (269ms)
        √ should report parsedSize and gzipSize when bundleDirOption is null (264ms)
        √ should report statSize only bundleDirOption is false (270ms)
```

### Related issues
- #274


th0r:
#274 is fixed by #275

PR #290, opened 2019-07-01
Add support for 'auto' option with analyzer port #281

### LINKED ISSUE #281: Convert `analyzerPort: "auto"` option to `analyzerPort: 0`
Currently to use automatic port selection one should use `analyzerPort: 0` option.
Despite the face that it's a value from a native Node.js API it's a little non intuitive.
We can introduce an alias: `analyzerPort: "auto"` and replace it with `0` before passing to http server.

TODO:
- Add alias
- Document it

avin-kavish:
It's a single line of quite self-explanatory code. But I'm happy to write tests, just not sure where.

The logic I went with was, if analyzer port is specified, (and) if it is 'auto' use 0, else use the specified value. If it is not specified, use 8888, the default.

Closes #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ new BundleAnalyzerPlugin(options?: object)
 |:--:|:--:|:----------|
 |**`analyzerMode`**|One of: `server`, `static`, `disabled`|Default: `server`. In `server` mode analyzer will start HTTP server to show bundle report. In `static` mode single HTML file with bundle report will be generated. In `disabled` mode you can use this plugin to just generate Webpack Stats JSON file by setting `generateStatsFile` to `true`. |
 |**`analyzerHost`**|`{String}`|Default: `127.0.0.1`. Host that will be used in `server` mode to start HTTP server.|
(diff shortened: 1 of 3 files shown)


avin-kavish:
`"before each" hook for "should support stats files with all the information in `children` array":`

Some issue with nightmare I think. Tests run fine on my pc

th0r:
Can we support `auto` value for CLI `--port` option as well?

th0r:
Could you please also update [this section in readme](https://github.com/webpack-contrib/webpack-bundle-analyzer#options)?

avin-kavish:
Sorry about the delay, I had some projects to work on. Not sure if you got the notification, I've force pushed to keep the commit history clean.

th0r:
Everything is fine except some issue with the build on CI. Could you take a look at why it fails?

th0r:
@avin-kavish Thanks for the PR, merged!

th0r:
Published in `v3.4.0`

PR #302, opened 2019-07-30
fix: solved the error — options is required

### LINKED ISSUE #300: options is required for 3.4.0
in the new 3.4.0,I found options of new BundleAnalyzerPlugin(options?: object) that it become required, otherwise it will get an error: Cannot use 'in' operator to search for 'analyzerPort' in undefined. if using the new BundleAnalyzerPlugin({}) in this way is successful.Is there a problem with 3.4.0？

jerryOnlyZRJ:
add a default value for BundleAnalyzerPlugin's options
fixes #300

### DIFF
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -8,7 +8,7 @@ const viewer = require('./viewer');
 
 class BundleAnalyzerPlugin {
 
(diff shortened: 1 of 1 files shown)


valscion:
Thanks!

jerryOnlyZRJ:
A piece of cake.

Issue #297, opened 2019-07-23
No Parsed/Gziped sizes if there is build hash in output path


martinkadlec0:
#### Issue description
If there as build hash in webpack's output path, I get following error from WBA:
`Error parsing bundle asset "/xxx/app/dist/[hash]/app.js": no such file`.

My webpack output config:
```
	output: {
		filename: '[name].js',
		path: path.join(__dirname, './dist/[hash]/'),
		publicPath: `${cdnPublicPath}[hash]/`,
		chunkFilename: '[name].js',
	},
```

This causes that there is no gziped/parsed size on the WBA page.

#### Technical info
* Webpack Bundle Analyzer version: 3.3.2
* Webpack version: 4.35.0
* Node.js version: 10.13
* npm/yarn version: npm 6.9
* OS: Mac OS Mojave

#### Debug info
How do you use this module? As CLI utility or as plugin?
Plugin

If plugin, what options were provided? (e.g. `new BundleAnalyzerPlugin({ analyzerMode: 'disabled', generateStatsFile: true })`)
No options

What other Webpack plugins were used?
webpack.DefinePlugin
CircularDependencyPlugin
FlowStatusWebpackPlugin
ExtractCssChunks
HtmlWebpackPlugin
SpriteLoaderPlugin

valscion:
Are you able to create a reproduction repository so that some contributor or us could check this error?

martinkadlec0:
Sure, there you go https://github.com/martinkadlec0/wba-build-hash

valscion:
Ah ok, now I get it! Thank you, the reproduction repository helped.

This does seem to be a lack of a feature that could also be categorized as a bug. The crux is this webpack config:

```
output: {
    filename: '[name].js',
    path: __dirname + '/dist/[hash]/',
  },
```

and that `webpack-bundle-analyzer` uses the `path` as-is instead of using some version where the `[hash]` would've been replaced with what it was in the end.

<img width="710" alt="vesa_Vesas-MacBook-Pro____code_muut_webpack-bundle-analyzer_wba-build-hash__zsh_" src="https://user-images.githubusercontent.com/482561/62151290-aa502480-b308-11e9-9a37-313d5c3efbed.png">
[...]

PR #304, opened 2019-07-30
Support output.path interpolation

valscion:
Resolves https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/297

### DIFF
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -35,6 +35,7 @@ class BundleAnalyzerPlugin {
 
     const done = (stats, callback) => {
       callback = callback || (() => {});
+      const bundleDir = stats.compilation.getPath(stats.compilation.outputOptions.path);
 
       const actions = [];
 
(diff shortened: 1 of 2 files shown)


src/BundleAnalyzerPlugin.js
```diff
@@ -35,6 +35,7 @@ class BundleAnalyzerPlugin {
 
     const done = (stats, callback) => {
       callback = callback || (() => {});
+      const bundleDir = stats.compilation.getPath(stats.compilation.outputOptions.path);
```
valscion: I have no idea if this `getPath` function is a public API from webpack. And neither do I know if `stats.compilation.outputOptions.path` is OK to use here.

test/plugin.js
```diff
@@ -47,6 +47,21 @@ describe('Plugin', function () {
     });
   });
 
+  it('should support webpack config with interpolation in `output.path`', async function () {
+    const config = makeWebpackConfig();
+    config.output.path += '/[hash]';
+
+    await webpackCompile(config);
+
+    const chartData = await getChartDataFromReport('7e56a1f8a54acedcc9ce/report.html');
```
valscion: I'm not sure if this hash is good across different webpack versions... this might cause a very brittle test.

I don't yet have good ideas on how to figure out the hash that was used to get to the report page.

src/BundleAnalyzerPlugin.js
```diff
@@ -48,9 +49,9 @@ class BundleAnalyzerPlugin {
       }
 
       if (this.opts.analyzerMode === 'server') {
-        actions.push(() => this.startAnalyzerServer(stats.toJson()));
+        actions.push(() => this.startAnalyzerServer(stats.toJson(), bundleDir));
       } else if (this.opts.analyzerMode === 'static') {
-        actions.push(() => this.generateStaticReport(stats.toJson()));
+        actions.push(() => this.generateStaticReport(stats.toJson(), bundleDir));
```
valscion: I'm not a huge fan in how I had to add another argument to these methods to get this working

PR #310, opened 2019-09-11
Update web page title

gaokun:
#### 1. add
add **favicon.ico** for web page.

#### 2. update
update web page title. 

Use `project name` + `build time` instead of `Webpack Bundle Analyzer`.

## Why & What scene

**Before:**

I would leave some 'Webpack Bundle Analyzer' tabs there, to see the difference between them.

> Like, the first tab is original. the second tab is packed with `tree-shaking`

image

But I often mix them up, coz same title.

**After:**

image

I use `project name` + `build time` as title that can help me to distinguish them.

### In one word:

**1. Icon represents this is a Webpack Bundle Analyzer page**

**2. Title represents which project it belongs to and the building time**

PS: Feel free to change a new icon you like.


valscion:
Nice idea! Could you split this to two pull requests — I'm not sure about favicon as the output is no longer one HTML file after this, but also contains the favicon. I like the idea of changing the title, though :+1:

valscion:
It might be possible to inject the favicon as a base64 encoded data URI, so that the output would still consist of only one HTML file.

package.json
```diff
@@ -64,6 +64,7 @@
     "chai": "4.2.0",
     "chai-subset": "1.6.0",
     "classnames": "2.2.6",
+    "copy-webpack-plugin": "^5.0.4",
```
th0r: Remove `^` please

package.json
```diff
@@ -77,7 +78,7 @@
     "gulp-babel": "8.0.0",
     "mobx": "5.9.4",
     "mobx-preact": "3.0.0",
-    "mocha": "6.1.2",
+    "mocha": "^6.2.0",
```
th0r: Why do you need to update `mocha` in this PR?

src/utils.js
```diff
@@ -32,3 +32,17 @@ function createAssetsFilter(excludePatterns) {
...
+ * format: MM/dd/yyyy HH:mm
+ * */
+exports.getCurrentTime = function () {
+  const time = new Date();
+  const year = time.getFullYear();
+  const month = time.getMonth() + 1;
+  const day = time.getDate();
+  const hour = time.getHours();
+  const minute = time.getMinutes();
+  return `${month}/${day}/${year} ${hour}:${minute}`;
```
th0r: The format worries me a little as e.g. I'm Russian and we use `day.month.year` format.
Shall we use some universal format like `21 Feb 2019`?

src/viewer.js
```diff
@@ -23,6 +24,8 @@ module.exports = {
   start: startServer
 };
 
+const title = `${process.env.npm_package_name || 'Webpack Bundle Analyzer'} [${utils.getCurrentTime()}]`;
```
th0r: Who sets `npm_package_name` env variable? Is it `npm`?
Does it mean that it will be empty if I run webpack via `gulp`?

views/viewer.ejs
```diff
@@ -3,7 +3,8 @@
   <head>
     <meta charset="UTF-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1"/>
-    <title>Webpack Bundle Analyzer</title>
+    <title><%- title %></title>
```
th0r: We should escape title:
```suggestion
    <title><%= title %></title>
```

views/viewer.ejs
```diff
@@ -3,7 +3,8 @@
...
+    <link rel="icon" href="favicon.ico">
```
th0r: Let's inline the icon using base64 encoding.

webpack.config.js
```diff
@@ -143,6 +144,14 @@ module.exports = opts => {
         );
       }
 
+      plugins.push(new CopyWebpackPlugin([
```
th0r: Let's better inline the icon as my wish is to keep the report to be a single html file.

src/viewer.js
gaokun: Yes, `npm` set env variable. 
We can use `npm scripts` to run webpack via `gulp`.

Or use `fs` to `readFile`? Feel kind of weird.
I think this is fine. We have default name after all.

gaokun:
@valscion glad to know u love it.

@th0r Hi, new commit pushed. Thanks for reviewing.

PR #314, opened 2019-09-22
Add leading zero to hour & minute on <title /> when needed

mhxbe:
Hi, 

Recently @gaokun requested/added a date in the <title /> tag which is a great addition to the tool.
Last night, when I was using webpack-bundle-analyzer on a personal project, I saw the following date:
```
<title>react-starter [21 Sep 2019 at 1:5]</title>
```

**Problem**
By default, `.getHours()` & `.getMinutes()` return a number which results in a one-digit for all numbers lower than 10. In this case, it's displaying **1:5** for the hour & minutes whereas **01:05** would be nicer. 

**Possible solution**
Adding a leading zero in combination with converting it to a String, leverages us the usage of `.slice(-2)` so it returns the last 2 digits. 

Another solution is the usage of .substr(-2) but it seems this method will be deprecated in the future
Source for `.slice()`: [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice)
Source for `.substr()`: [String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr)

If there are better approaches, I'm open for them.

**Examples**
You can copy paste the following code-snippet in your browser console to see it in action.
``` 
const someDate = new Date('2019-09-21 01:05');

console.log('Hours:', someDate.getHours());
console.log('Minutes:', someDate.getMinutes());

console.log('With leading zeros:');
console.log('Hours:', `0${someDate.getHours()}`.slice(-2));
console.log('Minutes:', `0${someDate.getMinutes()}`.slice(-2));
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,19 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
- * **Bug Fix**
+* **Bug Fix**
    * Add leading zero to hour & minute on `<title />` when needed ([#314](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/314) by [@mhxbe](https://github.com/mhxbe))
+
+## 3.6.0
+
+ * **Improvement**
+   * Support webpack builds where `output.globalObject` is set to `'self'` ([#323](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/323) by [@lemonmade](https://github.com/lemonmade))
+   * Improve readability of tooltips ([#320](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/320) by [@lorenzos](https://github.com/lorenzos))
+
+## 3.5.2
+
+ * **Bug Fix**
+   * Fix sidebar not showing visibility status of chunks hidden via popup menu (issue [#316](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/316) by [@gaokun](https://github.com/gaokun), fixed in [#317](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/317) by [@bregenspan](https://github.com/bregenspan))
 
 ## 3.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
+## 3.5.1
+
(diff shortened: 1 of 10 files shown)


gaokun:
Thanks for mention this. It is nice format to read.

And what about ES6 `padStart` function?

```js
'4'.padStart(2, 0);  //  result: '04'
```

In this case, they are both OK, anyway.

mhxbe:
That's even better and much cleaner, thanks for the suggestion.

I quickly checked the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart#Specifications) and they state that `.padStart()` is part of ES2017.
This means it's covered with `@babel/present-env` but there's a possible catch: IE11-support.

Is it safe to say this will work in both cases (server & client) since [src/viewer.js](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/src/viewer.js) is 'server-side' code thus not affecting IE11?

th0r:
We doesn't support IE11 in client bundle so it's completely safe to use padStart.

mhxbe:
Ok, thanks for the quick response @th0r.

I've pushed another commit where I used template literals to convert the number to a String.
```
const hour = `${time.getHours()}`.padStart(2,0);
const minute = `${time.getMinutes()}`.padStart(2,0);
```

I can change this to the following if preferred:
```
const hour = time.getHours().toString().padStart(2,0);
const minute = time.getMinutes().toString().padStart(2,0);
```

th0r:
Tag templates are ok, but let's better change method calls to `padStart(2, '0')` to be more clear.

th0r:
According to tests, `padStart` is not available in Node v6. 😕
I don't want to add another runtime dependency to polyfill it so let's better use `slice` approach.

valscion:
Hi @mhxbe, it would be great to get this bug fixed :relaxed:. Are you still up for this PR?

mhxbe:
Hi @valscion, I've been busy past week(s) but I will get this fixed tonight.
It's possible I'll reopen a new PR because I've made some 'polluted' commits. 

If you've got time to fix this now, feel free to open up a new PR with [the fix](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/314/commits/27b2de0d4a4334ccfdc2a44424984c563f39d06d). I'll happily close this one then. If this is not the case, I will fix this tonight.

valscion:
Ok this seems good to go to me! Thanks :relaxed: I'll merge once CI is green

PR #317, opened 2019-09-25
Fix case where checked bundles in sidebar don't update

### LINKED ISSUE #316: [Bug] Sidebar didn't show correct chunks when fixing.
#### Issue description
Fix the sidebar, then hide `app.js`, we can still see it as checked in sidebar.

image

#### Technical info
- Webpack Bundle Analyzer version: 3.5.0
- Webpack version: 4.39.3
- Node.js version: 10.15.3
- npm/yarn version: 6.11.3
- OS: MacOS

bregenspan:
Fixes #316

The relevant piece of state here is derived from props, and was failing to update upon prop update in the particular case noted in #316.

One thought re: preventing this class of bug in future is to centralize the state management logic that handles whether all bundles are showing, putting it in the shared MobX store (and eliminating use of state derived from props). I might have time to explore this in a separate PR, but I think this quick fix is probably better for now.

### Notes re: manual testing
Because the `componentWillReceiveProps` handler appears to have been added for <a href="https://github.com/bregenspan/webpack-bundle-analyzer/commit/279bb2638f1cddf04b319966c71565c48d874f3a">`webpack --watch`</a> support, I tested that the watch mode/dev server compatibility still functions as expected after this change by adding and removing bundles to a running webpack-dev-server. The checkbox state behaved as expected:

 * in the case where all checkboxes were checked before the new bundle was added, the new bundle is added in checked state
 * in the case where not all checkboxes were checked before the new bundle was added, the new bundle is not added in checked state

### DIFF
--- a/client/components/CheckboxList.jsx
+++ b/client/components/CheckboxList.jsx
@@ -33,6 +33,8 @@ export default class CheckboxList extends PureComponent {
         this.setState({checkedItems});
         this.informAboutChange(checkedItems);
       }
+    } else if (newProps.checkedItems !== this.props.checkedItems) {
+      this.setState({checkedItems: newProps.checkedItems});
     }
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
+ * **Bug Fix**
+   * Fix sidebar not showing visibility status of chunks hidden via popup menu (issue [#316](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/316) by [@gaokun](https://github.com/gaokun), fixed in [#317](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/317) by [@bregenspan](https://github.com/bregenspan))
+
 ## 3.5.1
 
  * **Bug Fix**


valscion:
@gaokun can you test out this PR to see if it fixes your problem in #317?

client/components/CheckboxList.jsx
```diff
@@ -33,6 +33,8 @@ export default class CheckboxList extends PureComponent {
         this.setState({checkedItems});
         this.informAboutChange(checkedItems);
       }
+    } else if (newProps.checkedItems !== this.props.checkedItems) {
```
bregenspan: Though I lean towards this being a good short-term fix, this method is now getting difficult to reason about IMO (in way that lines up with some of the arguments made <a href="https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html">here</a>). A few possibilities to simplify a bit, going from slight change to more radical change:

 1. Switch from `componentWillReceiveProps` (deprecated) to `getDerivedStateFromProps`. Would make intent clearer IMO, but might not be able to fully replace the `this.informAboutChange` part.
 2. Centralize the "all checked" logic in MobX store
 3. Use Hooks (could clean up a lot but biggest change / introduces new paradigm to the app)

(I like idea of trying option 2 personally)

gaokun:
> @gaokun can you test out this PR to see if it fixes your problem in #316?

Tested, it works now! ^_^

client/components/CheckboxList.jsx
valscion: I'll let @th0r weigh in on the refactoring opportunity :relaxed:

For now, as this change fixes the bug, let's ship this and consider the refactor in its own PR as it makes it easier to review a refactor when it keeps the existing functionality 1:1 to what was before.

valscion:
Released in v3.5.2 :tada:

Issue #321, opened 2019-10-09
Add a non-default option to open the report with "Show content of concatenated modules" already enabled


lorenzos:
Webpack does concatenate nearly all my source modules (React components), therefore I always need to open the sidebar and check "Show content of concatenated modules" when analyzing my bundles.

I see there are [good reason to not enable it by default](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/188), but it'd be great to have an opt-in flag in the options object to open the report with concatenated modules already shown.

valscion:
Hmm I'm not sure if that's worth the cost added by a new option and maintenance burden. I think it should always be an informed choice to show inaccurate information, so not adding it as an option would force people to always understand that the information they're seeing is indeed inaccurate.

The people writing the configuration can be different people to those who view the analysis, and that I'm worried about.

What do you think, @th0r?

th0r:
@valscion I wonder, maybe we can store such settings in `localStorage`? Won't it solve your problem @lorenzos?

valscion:
It could — that way the UI would remember your choice but you would've been the one to make that choice in the first place, not the one who wrote the configuration.

lorenzos:
@th0r Sure, it'll be ok for my use case.

valscion:
A PR to implement such `localStorage` option is welcome ☺️

PR #322, opened 2019-10-10
Persisted "Show content of concatenated modules" with localStorage

lorenzos:
As discussed in #321.

To my surprise, this works also when `{ analyzerMode: 'static' }`: I didn't know browsers save `localStorage` for local files too.

### DIFF
--- a/client/components/ModulesTreemap.jsx
+++ b/client/components/ModulesTreemap.jsx
@@ -208,6 +208,11 @@ export default class ModulesTreemap extends Component {
 
   handleConcatenatedModulesContentToggle = flag => {
     store.showConcatenatedModulesContent = flag;
+    if (flag) {
+      window.localStorage.setItem('showConcatenatedModulesContent', '1');
+    } else {
+      window.localStorage.removeItem('showConcatenatedModulesContent');
+    }
   }
 
   handleChunkContextMenuHide = () => {
--- a/client/components/ModulesTreemap.jsx
+++ b/client/components/ModulesTreemap.jsx
@@ -209,7 +209,7 @@ export default class ModulesTreemap extends Component {
   handleConcatenatedModulesContentToggle = flag => {
     store.showConcatenatedModulesContent = flag;
     if (flag) {
-      localStorage.setItem('showConcatenatedModulesContent', '1');
+      localStorage.setItem('showConcatenatedModulesContent', true);
     } else {
       localStorage.removeItem('showConcatenatedModulesContent');
     }
--- a/client/components/ModulesTreemap.jsx
+++ b/client/components/ModulesTreemap.jsx
@@ -4,7 +4,7 @@ import filesize from 'filesize';
 import {computed} from 'mobx';
 import {observer} from 'mobx-preact';
 
-import {isChunkParsed} from '../utils';
+import {isChunkParsed, localStorage} from '../utils';
 import Treemap from './Treemap';
 import Tooltip from './Tooltip';
 import Switcher from './Switcher';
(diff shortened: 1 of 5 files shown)


valscion:
Thanks! We should probably guard for `localStorage` methods raising errors, as there are plenty of situations where they could fail and the UI should work anyway

th0r:
Yep, we should create a small module that will wrap all the `localStorage` methods with `try {} catch`.

lorenzos:
Done in latest commit. Instead of creating a new module, I used `utils.js`.

client/utils.js
```diff
@@ -17,3 +17,27 @@ export function walkModules(modules, cb) {
 export function elementIsOutside(elem, container) {
   return !(elem === container || container.contains(elem));
 }
+
+export const localStorage = {
```
th0r: Extract it into a separate module please

client/utils.js
```diff
@@ -17,3 +17,27 @@ export function walkModules(modules, cb) {
...
+
+  getItem(k) {
+    try {
+      return window.localStorage.getItem(k);
```
th0r: Let's create a constant that will contain a prefix e.g. `wba` and prepend it to all keys. 
Let's also `JSON.stringify/parse` values so we can store any primitive value, not just string.

This is how I see this function:
```js
getItem(key) {
  try {
    return JSON.parse(window.localStorage.getItem(`${KEY_PREFIX}.${key}`));
  catch (err) {
    return null;
  }
}
```

client/utils.js
```diff
@@ -17,3 +17,27 @@ export function walkModules(modules, cb) {
...
+    } catch (x) {
+      return null;
+    }
+  },
+
+  setItem(k, v) {
```
th0r:
```suggestion
  setItem(key, value) {
```

client/utils.js
```diff
@@ -17,3 +17,27 @@ export function walkModules(modules, cb) {
...
+    try {
+      window.localStorage.setItem(k, v);
```
th0r:
```suggestion
      window.localStorage.setItem(key, JSON.stringify(value));
```

client/utils.js
```diff
@@ -17,3 +17,27 @@ export function walkModules(modules, cb) {
...
+    } catch (x) { /* ignored */ }
+  },
+
+  removeItem(k) {
```
th0r:
```suggestion
  removeItem(key) {
```

client/utils.js
```diff
@@ -17,3 +17,27 @@ export function walkModules(modules, cb) {
...
+    try {
+      window.localStorage.removeItem(k);
```
th0r:
```suggestion
      window.localStorage.removeItem(key);
```

client/components/ModulesTreemap.jsx
```diff
@@ -208,6 +208,11 @@ export default class ModulesTreemap extends Component {
 
   handleConcatenatedModulesContentToggle = flag => {
     store.showConcatenatedModulesContent = flag;
+    if (flag) {
+      localStorage.setItem('showConcatenatedModulesContent', '1');
```
th0r: Becomes this after implementing suggestions below:
```suggestion
      localStorage.setItem('showConcatenatedModulesContent', true);
```

lorenzos:
Should be ok now @ 9823e7c

valscion:
Looks good to me, what do you think @th0r?

It would be nice to have a changelog entry for this, similar to what we have for other things. This would be an "Improvement" change

lorenzos:
News on this?

th0r:
> It would be nice to have a changelog entry for this, similar to what we have for other things. This would be an "Improvement" change

@lorenzos

lorenzos:
> @lorenzos could you add a note to `CHANGELOG.md` under `UNRELEASED` section please?

Done. Sorry, I didn't get the request was addressed to me back then :)

th0r:
Thanks, merged!

valscion:
This has been released in v3.7.0 ☺️

Issue #288, opened 2019-06-23
Setting output.globalObject breaks parsed + gzip 


ntucker:
#### Issue description
```js
config.output.globalObject = "(typeof self !== 'undefined' ? self : this)"
```
Breaks parsed + gzip sizes. Removing it fixed them.

#### Technical info
* Webpack Bundle Analyzer version: 3.3.2
* Webpack version: 4.35.0
* Node.js version: 11.10.0
* npm/yarn version: 6.7.0
* OS: mac os x; windows

### Expected Behavior

Gzip and parsed should contain data bout more than just the output files - the modules inside them as well.

### Actual Behavior

Only stats includes all modules. Gzip and parsed just show large blocks for each bundle piece (output file)

### Code

My default:

```js
config.output.globalObject = "(typeof self !== 'undefined' ? self : this)"
```

When I analyze:
```js
delete config.output.globalObject;
```

valscion:
Sorry, I don't quite follow you. Is the piece of code you are referring to some part of `webpack-bundle-analyzer` or your own code?

ntucker:
the config setting lines are setting the webpack config. This issue describes what happens to any bundle when the configuration is as such.

valscion:
I'm having a hard time figuring out what is exactly what's happening here. Can you create a small reproduction repository to show what's happening?

I'm afraid I can't help you otherwise

ntucker:
https://webpack.js.org/configuration/output/#outputglobalobject This is the webpack configuration. You can use with any project. Just set that to  "(typeof self !== 'undefined' ? self : this)" and add bundle analyzer

valscion:
But why would you use `self` or `this` in there? Do you mean `'this'` instead of `this`?

ntucker:
https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/self

valscion:
Yes but the webpack configuration expects you to put a string in there, not an object, unless I misunderstand the instructions in https://webpack.js.org/configuration/output/#outputglobalobject documentation you linked?

ntucker:
"(typeof self !== 'undefined' ? self : this)" is a string literal. In javascript this is signified by the quotation marks surrounding the other characters.

valscion:
But it won't be evaluated as JS unless you put it in backticks? It means your string will be exactly what you write, not either `'self'` or `'this'`.

Was this what you have, or would need?

```js
[...]

PR #323, opened 2019-10-16
Allow self for globalObject in async chunks

lemonmade:
This PR adds support for `self` as the `globalObject` when searching for async chunks. We use `self` for our client bundles because it supports both browsers and web workers, and we have some libraries that will create workers as child compilations of the client compile. Technically you can pass a lot more in here (as shown in https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/288, although more complex configuration like that is still not supported with my change), but this PR just adds auto-detection of `self`. Given that it is technically the [default value](https://github.com/webpack/webpack/blob/0d4607c68e04a659fa58499e1332c97d5376368a/lib/WebpackOptionsDefaulter.js#L263), this seems like an appropriate inclusion to me.


test/bundles/validWebpack4AsyncChunkUsingSelfInsteadOfWindow.js
```diff
@@ -0,0 +1 @@
+(self.webpackJsonp=self.webpackJsonp||[]).push([[27],{1:function(e,n,t){console.log("Chuck Norris")}}]);
```
valscion: :smile:

valscion:
Thanks!

lemonmade:
❤️ Thanks so much for the quick review/ release!

PR #326, opened 2019-10-31
fix: fix worker assets not being included in report

### LINKED ISSUE #150: Doesn't parse child workerize'd module
#### Issue description
I use https://github.com/developit/workerize-loader to bundle some scripts for webworker, but bundle analizer doesn't parse and display it.

#### Technical info
* Webpack Bundle Analyzer version: 2.10.0
* Webpack version: 3.5.5.

image

As you see there is not `worker` module in the map:

image

nolanlawson:
Based on discussion [here](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/150#issuecomment-548406183), this should fix #150.

I've added a `stats.json` file from my own project [Pinafore](https://github.com/nolanlawson/pinafore) in order to test. I've confirmed that the test fails without the fix but succeeds with the fix, and that no other tests are broken.

Sorry for including such a large `stats.json` file in the test, but I wasn't sure I could come up with a minimal repro.

### DIFF
--- a/src/analyzer.js
+++ b/src/analyzer.js
@@ -36,7 +36,7 @@ function getViewerData(bundleStats, bundleDir, opts) {
     // See #22
     asset.name = asset.name.replace(FILENAME_QUERY_REGEXP, '');
 
-    return FILENAME_EXTENSIONS.test(asset.name) && !_.isEmpty(asset.chunks) && isAssetIncluded(asset.name);
+    return FILENAME_EXTENSIONS.test(asset.name) && isAssetIncluded(asset.name);
   });
 
   // Trying to parse bundle assets and get real module sizes if `bundleDir` is provided
--- a/test/analyzer.js
+++ b/test/analyzer.js
@@ -106,6 +106,14 @@ describe('Analyzer', function () {
     generateReportFrom('with-modules-chunk.json');
     await expectValidReport({bundleLabel: 'bundle.mjs'});
   });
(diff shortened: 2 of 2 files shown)


nolanlawson:
One of the odd things I observe with this PR, though, is that there is no chunk information for the worker assets – just the total size. I suppose this is better than nothing, but I'm not sure why Webpack doesn't seem to output chunk information for these assets?

Here is [a sample Webpack Bundle Analyzer file](https://bl.ocks.org/nolanlawson/raw/e77ed00a9cb9b925a140a210e7944477/) to demonstrate. Note that the asset named `blurhash.aaa4af0138af8a4c83aa.blurhash.js` contains no information other than its total size, even though Webpack built it from [this source file](https://github.com/nolanlawson/pinafore/blob/ee492c17653a9e900814f4005c4853a1d67b9ea9/src/routes/_workers/blurhash.js), so you'd think it would have chunk information.

valscion:
Hi! It seems that I've forgotten about this PR, sorry.

> Sorry for including such a large `stats.json` file in the test, but I wasn't sure I could come up with a minimal repro.

Unfortunately this will make it impossible for us to merge this PR. We would really like to have as minimal as possible `stats.json` files.

Would you be interested in trying to come up with such a change or should we close this PR?

th0r:
Closing because of inactivity.

PR #338, opened 2020-03-09
Upgrade acorn to v7.1.1

mulholio:
Upgrades acorn and acorn-walk to 7.1.1 in order to avoid the NPM audit warning here: https://www.npmjs.com/advisories/1488.

---

Tests all pass fine but I am getting these warnings from `npm` if I update `acorn` and `acorn-walk` only.

```
npm WARN acorn-dynamic-import@4.0.0 requires a peer of acorn@^6.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN acorn-jsx@5.0.1 requires a peer of acorn@^6.0.0 but none is installed. You must install peer dependencies yourself.
```

`acorn-jsx` is relied upon by `eslint`. `acorn-dynamic-import` is relied upon by `webpack`.

For `acorn-jsx`, I wanted to upgrade `eslint` but all this does is remove the one `npm WARN` above and replace it with three others, including one from `eslint-config-th0r` and one from `eslint-config-th0r-react` which haven't been updated in a year and rely upon `eslint` v5.x.x. It seems out of scope to change these packages upstream.

I have not yet attempted to update `webpack` but that also feels excessive here.

What is `webpack-bundle-analyzer`'s policy on having `npm WARN`s? Are they fine to have here? If not, do the maintainers have suggestions on how to work around the above warnings?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1172,9 +1172,9 @@
       }
     },
     "acorn": {
-      "version": "6.2.1",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-6.2.1.tgz",
-      "integrity": "sha512-JD0xT5FCRDNyjDda3Lrg/IxFscp9q4tiYtxE1/nOzlKCk7hIRuYjhq1kCNkbPjMRMZuFq20HNQn1I9k8Oj0E+Q=="
+      "version": "7.1.1",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-7.1.1.tgz",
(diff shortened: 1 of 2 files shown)


th0r:
@mulholio those warnings are not good - packages expect another versions of `acorn` so we can't just ignore it.

mulholio:
> @mulholio those warnings are not good - packages expect another versions of acorn so we can't just ignore it.

Would you be able to help in upgrading the lint rules package so that it expects a newer version of eslint?

th0r:
Actually, I don't see these unmet peer dependency warnings. Was able to update a few packages in #339 and get rid of vulnerability warnings completely. Thanks for the PR!

Issue #194, opened 2018-07-04
Feature request: Generate analysis report file


jasperck:
#### Issue description
Thanks for the awesome plugin, it truly helps us analyzing how our bundles had been built. Recently, I'm looking to analyze size of bundles in background or on CI. `webpack-bundle-analyzer` does great help for creating and storing the `stats` file from `webpack`, which gives me the opportunity to tweak the data in file. But `webpack-bundle-analyzer` only store the origin stats file, if we in some cases need information like `gzip size`, then there's no chance here.

In this case, the data which `report.html` used seems to be more valuable. Thus I think it would be awesome to make `webpack-bundle-analyzer` possible to generates and stores the data.

```
# Plugin options could be
{
  generateAnalysisFile: false // Boolean
  analysisFilename: 'analysis.json' // string
}
```

```
# CLI options could be
-a, --analysis <file>  Path to bundle analysis file that will be generated. (default: analysis.json)
```

Please feel free to close this if it's not something this plugin should response for. Otherwise, I'd be happy to create a PR for this. Also, suggestions and discussion are super welcome 🙏😀

jasperck:
Hi @valscion, sorry it's been such a long time since I drafted this issue.
I just created a PR https://github.com/jasperck/react-gobang/pull/3/files for demo what report file could be generated from the `webpack-bundle-analyzer` that I stated. The PR shows 2 things,
1. Usage
It just extend with a `json` mode, which tells `webpack-bundle-analyzer` generate `report.json` than `report.html`, and we could have the valuable analysis file.
2. Report file
With kind of data, we could do something like `group/filter/compare/...etc` and take advantage of it not only for visualize it but also for other workflow benefit.
```
[
  {
    "label": "static/js/main.2bd70c8a.js",
    "statSize": 744348,
    "parsedSize": 358496,
    "gzipSize": 110526,
    "groups": [
      {
        "label": "node_modules",
        "path": "./node_modules",
        "statSize": 713880,
        "groups": [
          {
            "label": "react",
            "path": "./node_modules/react",
            "statSize": 7021,
            "groups": [
              {
                "id": 0,
                "label": "index.js",
...
```
## Example
We integrated this with `CircleCI` and `DangerJS`, and now we can view bundle size changes within each PR, and all these were done automatically
[...]

PR #341, opened 2020-04-03
Added JSON output to the cli and webpack plugin

Gongreg:
## Issue
This issue can be related to https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/194.

We would like to use the statistics generated by Webpack Bundle Analyzer to track changes in projects. For that we would like to upload/save the bundle sizes in a database, and json is a lot more flexible format for that. 

## Tests

I've added a test for the json output.

## Next step

There is already possibility to generate html from the given json (you expose ejs file that we can directly use). But I think it would be great to add a way to generate the html from cli, but I can't think of a nice way to do it. This is what I currently have created: https://github.com/Gongreg/webpack-bundle-analyzer/pull/1/files

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -133,14 +133,15 @@ async function generateReport(bundleStats, opts) {
 
   if (!chartData) return;
 
-  function saveReport(content) {
-    const reportFilenameWithExtension = reportFilename
-      ? reportFilename
-      : reportFormat === 'static'
-        ? 'report.html'
-        : 'report.json';
+  const reportFilenameWithExtension = reportFilename
+    ? reportFilename
+    : reportFormat === 'static'
+      ? 'report.html'
+      : 'report.json';
 
-    const reportFilepath = path.resolve(reportFilenameWithExtension);
+  const reportFilepath = path.resolve(reportFilenameWithExtension);
+
+  function saveReport(content) {
     mkdir.sync(path.dirname(reportFilepath));
     fs.writeFileSync(reportFilepath, content);
 
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -183,7 +183,7 @@ async function generateJSONReport(bundleStats, opts) {
   mkdir.sync(path.dirname(reportFilename));
   fs.writeFileSync(reportFilename, JSON.stringify(chartData));
 
-  logger.info(`${bold('Webpack Bundle Analyzer')} saved report to ${bold(reportFilename)}`);
+  logger.info(`${bold('Webpack Bundle Analyzer')} saved JSON report to ${bold(reportFilename)}`);
 }
 
 function getAssetContent(filename) {
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -133,8 +133,6 @@ async function generateReport(bundleStats, opts) {
 
   if (!chartData) return;
 
-  const reportFilepath = path.resolve(reportFilename);
-
   await new Promise((resolve, reject) => {
     ejs.renderFile(
       `${projectRoot}/views/viewer.ejs`,
@@ -156,13 +154,13 @@ async function generateReport(bundleStats, opts) {
             return;
           }
 
-          mkdir.sync(path.dirname(reportFilepath));
-          fs.writeFileSync(reportFilepath, reportHtml);
+          mkdir.sync(path.dirname(reportFilename));
+          fs.writeFileSync(reportFilename, reportHtml);
 
-          logger.info(`${bold('Webpack Bundle Analyzer')} saved report to ${bold(reportFilepath)}`);
+          logger.info(`${bold('Webpack Bundle Analyzer')} saved report to ${bold(reportFilename)}`);
 
           if (openBrowser) {
-            opener(`file://${reportFilepath}`);
+            opener(`file://${reportFilename}`);
           }
           resolve();
         } catch (e) {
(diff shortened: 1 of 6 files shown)


valscion:
Hi, thanks for opening a pull request for this :relaxed:. Gotta say that this is a surprisingly small changeset! I expected this to need more code.

I wonder if @th0r has opinions about letting people save the generated chart data as JSON instead of HTML with a new option?

This approach taken here seems quite legit to me!

Gongreg:
Glad to hear that!

If you will decide to include this functionality and there are any changes needed for it please let me know :)

valscion:
Will do :)

src/viewer.js
```diff
@@ -153,14 +175,7 @@ async function generateReport(bundleStats, opts) {
             return;
           }
 
-          const reportFilepath = path.resolve(bundleDir || process.cwd(), reportFilename);
```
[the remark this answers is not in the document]
Gongreg: Seems like there were few possible cases:

1. Manually pass in report name to Plugin config/CLI options, then the cli/plugin resolves the full path before passing it to the `generateReport`, and `path.resolve(bundleDir || process.cwd(), reportFilename);` would always simply return `reportFilename`.

2. You do not pass report name to Plugin config/CLI options. They default to `report.html` and the path is also resolved before calling `generateReport`. Same happens as in case 1.

So overall path.resolve and || inside `generateReport` were redundant. Unless I am missing something :)

Since I am splitting out json/html generation to two separate functions I will revert back path resolvement to outside of `generateReport`, to make sure that I am not changing existing functionality.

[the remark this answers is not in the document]
valscion: Thanks! Seems like indeed it might be redundant, but better play it safe and not change it here :smile:

Gongreg:
Hey @valscion. I've updated the PR according to the comments. Wrote one reply.

I wanted to ask about node api (I mean the exposed methods in the entry of the package). Could we also expose the generateReport/generateJSONReport methods there? It would simplify using this package inside other node projects.

Also that would allow to implement generation of html from json in a proper way instead of abusing the cli (as I showed here: https://github.com/Gongreg/webpack-bundle-analyzer/pull/1/files)

valscion:
I'm not yet sold on the idea of exposing the generation of HTML from JSON so I'd rather not delve too deep into that subject here, if that's okay to you ☺️.

Let's get the JSON output format out the door first.

> I wanted to ask about node api (I mean the exposed methods in the entry of the package). At the moment the method to start the server is exposed.

Sorry, I'm running a bit slow today — which file are you referring to here?

Gongreg:
https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/src/index.js

valscion:
> At the moment the method to start the server is exposed.

I think that's legacy API and we might even want to get rid of it. We haven't documented its existence in the README. It's been there since I joined maintaining this plugin, so I don't know what the idea behind it is.

I'd like to avoid adding any new top-level APIs here outside of the new analyzer mode option. We can of course open a discussion about the top-level API — that discussion would preferably be outside this PR, though, in its own issue ☺️

src/viewer.js
```diff
@@ -153,17 +154,13 @@ async function generateReport(bundleStats, opts) {
...
-          fs.writeFileSync(reportFilepath, reportHtml);
-
-          logger.info(
-            `${bold('Webpack Bundle Analyzer')} saved report to ${bold(reportFilepath)}`
-          );
+          logger.info(`${bold('Webpack Bundle Analyzer')} saved report to ${bold(reportFilename)}`);
 
           if (openBrowser) {
-            opener(`file://${reportFilepath}`);
+            opener(`file://${reportFilename}`);
```
valscion, line 163: As per the other discussion we had https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/341/files#r404064871, can we avoid doing refactoring in this method in this PR apart from removing the `report.html` default value for `reportFilename` option?

I don't trust my reviewer eyes today enough to understand the code flow of this method, and would like to avoid changing any code here that we don't have to ☺️

Gongreg: Sure, but can I add a comment that it is most likely redundant? Just makes understand the code flow more difficult :)

Gongreg:
Sure,
I am trying to think of a way to nicely expose both json generation functionality (seems like this PR is good enough) and html report generation from the json. This is not a common use case and 99% of the users will not use it, but for those who wish to do the generation in multiple steps (store json somewhere and render when it is needed) it is important to be able to do it.

Should I create a separate issue for it? For now we will just directly render the ejs you have in the package.

valscion:
> Should I create a separate issue for it?

If you want to continue the conversation about this feature then sure ☺️. You can also open a separate pull request after this PR with your example change, and use that as a discussion starting point — we make no promises to merge that PR but it does help to have some code about a use case already available when discussing it ☺️

src/viewer.js
Gongreg: Reverted since it is not inside the scope of the current change. But I didn't add the path resolvement in `generateJSONReport`, since it is redundant.

src/viewer.js
```diff
@@ -174,6 +173,19 @@ async function generateReport(bundleStats, opts) {
...
+  const {reportFilename, bundleDir = null, logger = new Logger(), excludeAssets = null} = opts || {};
+
+  const chartData = getChartData({logger, excludeAssets}, bundleStats, bundleDir);
+
+  if (!chartData) return;
+
+  mkdir.sync(path.dirname(reportFilename));
+  fs.writeFileSync(reportFilename, JSON.stringify(chartData));
+
+  logger.info(`${bold('Webpack Bundle Analyzer')} saved report to ${bold(reportFilename)}`);
```
valscion, line 186: What do you think about showing that we saved a JSON report here?

```suggestion
  logger.info(`${bold('Webpack Bundle Analyzer')} saved JSON report to ${bold(reportFilename)}`);
```

I don't know if it's necessary but might be a nice additional touch ☺️

Gongreg:
Hey @valscion , any news? :)

valscion:
Thanks for the contribution! I'll hopefully remember to make a release next time I'm on my coding laptop ☺️. Might be 'til end of easter, Tuesday, but might be earlier. It depends on if I want to do some coding stuff during this long weekend or not 😄

Feel free to ping me if I haven't made a release by Wednesday ☺️

valscion:
This has been released in v3.7.0 ☺️

PR #345, opened 2020-04-24
add support for node bundle

Spikef:
when webpack config target=node, the modules are store at exports.modules.

https://github.com/webpack/webpack/blob/157c457241bf46b1cb31dbb17e806702eb4a0ec5/lib/node/NodeTemplatePlugin.js#L39


test/bundles/validNodeBundleWithModulesAsArray.js
```diff
@@ -0,0 +1,6 @@
+exports.ids = ["common"];
+exports.modules = {
```
th0r, line 2: Filename is "validNodeBundle**WithModulesAsArray**.js" but modules are in the object. Shall we remove `WithModulesAsArray` part?

src/parseUtils.js
```diff
@@ -25,6 +25,21 @@ function parseBundle(bundlePath) {
     ast,
     walkState,
     {
+      AssignmentExpression(node, state,) {
```
th0r, line 28:
```suggestion
      AssignmentExpression(node, state) {
```

src/parseUtils.js
```diff
@@ -25,6 +25,21 @@ function parseBundle(bundlePath) {
...
+        if (state.locations) return;
+
+        // Modules are stored in exports.modules:
+        // exports.modules = {};
+        const {left, right} = node;
+
+        if (
+          left &&
+          left.object && left.object.name === 'exports' &&
+          left.property && left.property.name === 'modules'
```
th0r, line 38: Lets also add `&& isModulesHash(right)` check.

th0r:
Merged, thanks!

valscion:
This has been released in v3.8.0 :tada:

PR #346, opened 2020-04-29
feat: Render Selected Module Size Total 

vwb:
Small UI update to render a label of the total selected module size in the sidebar. If the user has all the modules selected, or only one, it will not display as that information is easily accessible.

Image 4-28-20 at 6 31 PM

Have found myself needing this more often than not to compare the sizes of subsets of bundles.

Wasn't able to find any tests on UI components. If I'm simply blind, please point me in the right direction and I'm happy to add some coverage. 

Thanks!

### DIFF
--- a/client/components/ModulesTreemap.jsx
+++ b/client/components/ModulesTreemap.jsx
@@ -115,6 +115,11 @@ export default class ModulesTreemap extends Component {
                 onChange={this.handleSelectedChunksChange}/>
             </div>
           }
+          {this.isTotalSelectedSizeVisible() &&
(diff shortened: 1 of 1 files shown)


vwb:
Something seems to have gone wrong with the license bot, I completed the check months ago but still is saying pending. 

If there anything I can do on my side to address this let me know.

valscion:
I think I triggered a new CLA check with close+reopen. It looks like the commit in this PR is associated with https://github.com/vincent3 GitHub user account, and that user account might not have signed the CLA.

PR #348, opened 2020-05-03
feat: allows deterministic reports

eoingroat:
Allows building reports deterministically; a la [Deterministic Compilation / Reproducible Builds](https://en.wikipedia.org/wiki/Reproducible_builds)

For now, that consists of adding an option to remove the report generation time from the report title; making the report generation reproducible across time.

## tests

Tests are included, they check builds are deterministic with the deterministic API option true, and non-deterministic with the option false.

## why

I've adopted a zero-config (a la yarn 2) and fully reproducible artifact build and test strategy; which means I need to commit my reports into a form of version control.

## thoughts

I was going to propose this as a default to deterministic, however it may be the case that users use the included timestamp to determine the recency of the report when developing, and thus defaulted to the existing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
+ * **New Feature**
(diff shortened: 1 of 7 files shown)


eoingroat:
Committed with wrong account

PR #349, opened 2020-05-03
feat: Allows deterministic builds

eoingroat:
Allows building reports deterministically; a la [Deterministic Compilation / Reproducible Builds](https://en.wikipedia.org/wiki/Reproducible_builds)

For now, that consists of adding an option to remove the report generation time from the report title; making the report generation reproducible across time.

## tests

Tests are included, they check builds are deterministic with the deterministic API option true, and non-deterministic with the option false.

## why

I've adopted a zero-config (a la yarn 2) and fully reproducible artifact build and test strategy; which means I need to commit my reports into a form of version control.

## thoughts

I was going to propose this as a default to deterministic, however it may be the case that users use the included timestamp to determine the recency of the report when developing, and thus defaulted to the existing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
+ * **New Feature**
+   * Added deterministic option (`deterministic: true` in plugin, `--deterministic` in CLI) (https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/349) by [@eoingroat](https://github.com/eoingroat)
+
 ## 3.7.0
 
  * **New Feature**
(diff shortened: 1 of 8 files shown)


valscion:
Hi, thanks for the PR.

I'm not yet really into adding a configuration option for determinism, but maybe we could add a configuration option for the page title instead?

eoingroat:
@valscion I'm happy with that; I'll get to it sometime this week.

Would you be happy if I added a test to check that a deterministic build is possible through some configuration - either in the same or another PR?

eoingroat:
Closing in favour of #354

valscion:
> Would you be happy if I added a test to check that a deterministic build is possible through some configuration

I'm not yet sure if deterministic builds are ones we want to support as a first-class citizen, so preferably let's add that test in another PR later on so we can think on that separately from the title option ☺️

PR #352, opened 2020-05-19
Removed left side check to support custom global object which can be any string

### LINKED ISSUE #350: isAsyncChunkPushExpression is checking for left side which cannot be easily determined
#### Issue description

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/757c12e03b8260b3428db1c618224a90b69569c1/src/parseUtils.js#L191-L214

These lines https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/757c12e03b8260b3428db1c618224a90b69569c1/src/parseUtils.js#L201-L208

Are checking if .push is called on `window`, `self` or `this`. 
So for the simple cases:

```javascript
window.webpackJsonp_someCustomName = (window.webpackJsonp_someCustomName || []).push(...)
```
it works well.

But sometimes Webpack creates left side output which can be `MemberExpression`:

```javascript
(("undefined" != typeof self ? self : this).webpackJsonp_someCustomName = ("undefined" != typeof self ? self : this).webpackJsonp_someCustomName || []).push(...);
```

I am thinking is there any need for checking of the left side? We are already checking a lot of things and

```javascript
mayBeAsyncChunkArguments(args[0].elements) && isModulesList(args[0].elements[1])
```

should make us confident that it is indeed a module.

Gongreg:
Issue: #350 .

Removed left hand checks, since they seem redundant. All the tests are passing. Added test case for a custom global object.

Closes #350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 <!-- Add changelog entries for new changes under this section -->
 
  * **Bug Fix**
-   * Support any custom `globalObject` option in Webpack Config.
+   * Support any custom `globalObject` option in Webpack Config. ([#352](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/352) by [@Gongreg](https://github.com/Gongreg))
 
 ## 3.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
+ * **Bug Fix**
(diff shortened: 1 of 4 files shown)


valscion:
Thanks! Can you add a changelog entry for this to `CHANGELOG.md`? This seems like either an improvement or a bug fix, given that we now support parsing more bundles than before.

Gongreg:
done

valscion:
Thanks! This looks good to me. Now that I can see the full conditional, it's unlikely that this would suddenly start parsing "successfully" some file in the bundle output directory which would not be a valid webpack bundle.

Gongreg:
Yes, the right hand side check seems quite foul proof, on top of that this check is checking for `.push`.

valscion:
This has been released in v3.8.0 :tada:

PR #354, opened 2020-05-31
feat: reportTitle option (allow reproducible build)

eoingroat:
Allows setting the HTML report title

## tests

Tests are included, they test string and function reportTitle set through the plugin API, and error propagation for the function reportTitle

## why

I, and increasingly others, require reproducible builds with CI; this is a more specific PR targeting a single option after comments on a more general PR #349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 <!-- Add changelog entries for new changes under this section -->
 
+* **New Feature**
+   * Adds option reportTitle to set title in HTML reports; default remains date of report generation ([#354](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/354)) by [@eoingroat](https://github.com/eoingroat)
+
 ## 3.8.0
 
  * **Improvement**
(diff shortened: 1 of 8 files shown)


valscion:
Yeah this is looking much better, thanks!

It seems that the CLI currently requires the `--title` option and doesn't default to the previous title, right? Could we add a test for the CLI to make sure default title works? And maybe another test to test that the custom title also works.

I'm not sure if we have test coverage for the old title, either, so if you're in the mood, having some OK-level coverage for the existing title generation logic would be nice (i.e. if we can test the title generation without having to stub time, it would be nice — a regex-test on the title would also suffice)

eoingroat:
Ah yes; that was fairly low quality of me to not include those tests.

I've gone for a very vague regex for the title, mainly to eliminate potentially stringified error values

valscion:
Looking much better!

Now the default title is duplicated, and I wonder if we could de-duplicate and put it into `utils`. It seems that the `getCurrentTime` export is only used in generating the default title, so we can stop exporting `getCurrentTime` and instead move to export `getDefaultReportTitle` function which both the plugin-side and the CLI-side can use.

Also for the regex, I think we can be a bit more strict and hopefully make the regex a bit more clear. What do you think of using this regex?

```js
expect(generatedReportTitle).to.match(/^webpack-bundle-analyzer \[.* at \d{2}:\d{2}\]/u);
```

I'm not sure if there's a situation where `process.env.npm_package_name` would not be defined as `webpack-bundle-analyzer` when running the tests.

If however we would not like to rely on `webpack-bundle-analyzer` to be there, we could also be even more relaxed in the regex so that it would not be so brittle yet it would make sure that the time of day is printed in proper format:

```js
expect(generatedReportTitle).to.match(/^.+ \[.* at \d{2}:\d{2}\]/u);
```

PR #357, opened 2020-06-09
chore: do not throw if an invalid log-level is passed

jamesgeorge007:
A suitable message shows up along with the `help information` just as we have for `--default-sizes` flag.


valscion:
Thanks for the PR. However, this logger is also used in the plugin side, not only from the CLI.

As the CLI exits immediately after `showHelp` is called, I don't think this approach would be something we want. We do want the plugin-side to crash if an invalid log option is given.

jamesgeorge007:
Gotcha 👍

PR #360, opened 2020-06-11
update readme with npm shorter commands

sarkiroka:
no one types `npm install --save-dev...` but rather `npm i -D...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 ```bash
 # NPM
-npm install --save-dev webpack-bundle-analyzer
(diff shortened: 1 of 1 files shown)


valscion:
Thanks but we prefer the longer form.

sarkiroka:
@valscion could you explain why you prefer the longer form in npm usage, but not in yarn? In yarn usage you don't use the longer form: `yarn add --dev ...` instead of `yarn add -D ...`
I don't understand.

PR #374, opened 2020-08-07
Update package.json

RamyaPayyavula:
older versions of lodash and babel has lodash vulnerabilities


th0r:
Fixed in https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/389

PR #376, opened 2020-08-19
Allowing capability to parse bundles that have child assets generated, along with main asset

masterkidan:
This essentially attempts to solve #150, when worker-loader is used it generates bundles with the following compilation stat

1 main asset
Children Assets 
image

The main chunk that is emitted bundle.worker.js, doesn't have any children associated with, the webpack compiler instead spits out the dependencies as part of stats.children.assets. 

This PR attempts to solve the above by traversing the children assets and constructing trees for them as well.


masterkidan:
@jsf-clabot rerun

valscion:
Seems like the first commit in this PR has an email address associated with it that is not associated with your GitHub account. You will need to fix that e.g. with interactive rebasing and resetting the commit author details, and then force pushing your changed branch here to fix CLA check.

masterkidan:
@valscion : I've fixed the cla issues now, please review the changeset and lmk if it looks ok.

valscion:
I do appreciate the tests you added. I'm unfortunately not so knowledgeable in the specific codepath you're changing, so I'll defer to @th0r for reviewing.

Is there something in particular in this diff that you could do differently @masterkidan or that you think could cause issues in the future? Is there something that you think could cause maintenance burden?

masterkidan:
Nothing springs to my mind here, let’s wait for @th0r to review.

masterkidan:
@th0r: A gentle ping for a review, I would like to get this in so that I can unblock some local debugging issues am having with large bundle sizes for workers.

valscion:
Oh well, we can always fix any problems that arise if this code breaks. We also have a test case for this, and the code seems quite OK to me :relaxed:.

Let's move ahead with this one. Thank you for the contribution, @masterkidan ☺️

masterkidan:
Thanks @valscion: Is it possible to get a patch/minor release with the above changes so that I can consume the updates? Will it get generated automatically?

valscion:
Yeah I was in the progress of releasing a new version yesterday but then my time ran out. It's been a while since there was an update, so I'm looking at if I can also bump some dependencies in a backwards compatible manner before I release a new version.

I'll try to remember to comment here when we release a new version. I can't promise any given timeline, as it depends on when I have a nice time slot to do the update.

dabbott:
I was having some large worker bundle issues, so I ended up pulling master and trying this out.

I ran into an issue where an error is thrown because `getChildAssetBundles` returns undefined. I think it's potentially due to the `HtmlWebpackPlugin` in my project, but I really don't know the internals of webpack or bundle analyzer at all so I'm not sure. I found that for my asset names, e.g. `9.typescript-worker-bundle.js`, the only `chunkByAssetName` keys that existed were `[ 'HtmlWebpackPlugin_0' , 'main', 'main' ]`. That's as far as I got in debugging, but I'm afraid I don't have time at the moment to dig deeper.

Here's a quick fix if you want one: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/378

With that, this works great and I'm able to see worker bundle stats now 😃

valscion:
Thanks for testing these changes, @dabbott!

Issue #22, opened 2016-11-25
Cannot read property 'parsedSize'


konstantin24121:
Error in script.
Console said © Cannot read property 'parsedSize' of undefined

850: r.hasParsedSizes = "number" == typeof r.props.data[0].parsedSize,

th0r:
What does `window.chartData` variable in generated report contain? Is it empty?

konstantin24121:
totaly empty http://joxi.ru/xAeOogEiyZNkmy

th0r:
Well, then I need all the information from issue template that you erased. It'll simply finding a bug. At least it would be nice to see your stats file.

konstantin24121:
https://www.dropbox.com/sh/599ueqfbxq4715n/AACJlV949_dSR2C1m2lXFCtla?dl=0 my webpack config and two generated by WBA files

th0r:
You use very strange template for bundle filenames: `filename: '[name].js?v=[hash]'`.
What webpack generates in this case? Does your generated bundle files have `?v=[hash]` in the name (e.g. `bundle.js?v=8c80fe52746c2c5eb82e`)?

th0r:
The thing is analyzer filters out all non-js assets and your js bundles are filtered out as well because they don't have `.js` extension.

th0r:
Should be fixed in `v2.0.1`

konstantin24121:
@th0r  thanks!

PR #378, opened 2020-08-29
Fix missing child bundles throwing an error

dabbott:
This fixes an issue introduced in https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/376, where `getChildAssetBundles` sometimes doesn't find the given `statAsset.name` and returns undefined, which later causes an error.

This probably isn't a great fix, and I'm not sure specifically how to reproduce the issue, but it seems harmless enough that I'll be using it for now at least 😅


dabbott:
@jsf-clabot rerun

valscion:
Thanks! Would you be able to setup a test with `stats.json` that breaks, similar to files in here? https://github.com/webpack-contrib/webpack-bundle-analyzer/tree/master/test/stats

Those files are used by tests defined in https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/test/analyzer.js

th0r (changes requested):
@valscion let's not merge it without a test please.

valscion:
> @valscion let's not merge it without a test please.

:+1: sounds good to me!

masterkidan:
@dabbott : It would be great if you can debug a bit deeper, my understanding is that we eliminate assets generated by plugins like HtmlWebpackPlugin in the regex/assetfilter (only js assets are analyzed). It looks like in your webpack config something with HtmlWebpackPlugin is generating js assets, can you share the config so I can understand the issue a bit better?

dabbott:
@masterkidan Thanks for checking this out. Here's my webpack config: https://github.com/dabbott/react-native-web-player/blob/58e49839516a32a91267510b550823e644b72364/webpack/webpack.config.js. 

If you'd like, you should be able to pull down that `react-native-web-player/webpack-bundle-analyzer` branch and run `yarn build` to see the error (if you have `webpack-bundle-analyzer` linked locally so you're pointing to master). I don't know how webpack assets are organized _at all_ really, so I probably won't be much help in making a better fix, but if you have suggestions or a fix to test I can definitely try it out.

masterkidan:
@dabbott : I think this may be because of the fact your template is an 'ejs' file. This is not something that gets filtered out by the regex, but the template compilation is a child compilation, I think a slightly better way would be to check why
// Picking only `*.js or *.mjs` assets from bundle that has non-empty `chunks` array
  bundleStats.assets = _.filter(bundleStats.assets, asset => {
    // Removing query part from filename (yes, somebody uses it for some reason and Webpack supports it)
    // See #22
    asset.name = asset.name.replace(FILENAME_QUERY_REGEXP, '');

    return FILENAME_EXTENSIONS.test(asset.name) && !_.isEmpty(asset.chunks) && isAssetIncluded(asset.name);
  });
is not filtering out the asset, and fix it here instead :) . 

You can use https://github.com/masterkidan/worker-loader-no-stats as a base and add a simple index.ejs template with a WebpackHtlmPlugin to process it.... I am not sure when I will be able to get to it on my own.

dabbott:
Ok! I was able to track this down. This happens when workers contain dynamic imports with `import('./some-dep.js')`. I added a test case (generated via https://github.com/dabbott/worker-loader-dynamic-imports).

I think this is probably in an OK state to merge, and @masterkidan if you have time you could do a better fix in the future.

th0r:
@dabbott so, what is the actual reason for the exception? What asset it tries to find but it's not there?

dabbott:
We're looking for `"1.bundle.worker.js"` as `assetName` in:

```
function getChildAssetBundles(bundleStats, assetName) {
  return _.find(bundleStats.children, (c) =>
    _(c.assetsByChunkName)
      .values()
      .flatten()
      .includes(assetName)
  );
}
```

That name doesn't exist in `assetsByChunkName`. (Maybe cause it's not a named chunk? It does exist in `c.chunks` I believe)

Issue #377, opened 2020-08-27
Feature Request: Analyze gzipped bundles


RedVelocity:
#### Issue description
webpack-bundle-analyzer currently displays a blank page when a webpack build uses compression-webpack-plugin to compress bundles.

#### Technical info
Using the following workaround. This skips compression when webpack build command is used with 'analyze' environment variable.

```
module.exports = ({ env, analyze }) => {
  switch (env) {
    case 'development':
      return merge(commonConfig, developmentConfig);
    case 'production':
      if (analyze) commonConfig.plugins.push(new BundleAnalyzerPlugin());
      else
        productionConfig.plugins.push(
          new CompressionPlugin({
            filename: '[path].gz',
            algorithm: 'gzip',
            test: /\.(js|jsx|css|html|png|svg|jpg|gif)$/,
            threshold: 10240,
            minRatio: 0.8,
            deleteOriginalAssets: true,
          })
        );
      return merge(commonConfig, productionConfig);
    default:
      throw new Error('No matching configuration was found!');
  }
};
```

<pre>
  System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 8.06 GB / 15.81 GB
  Binaries:
    Node: 12.18.1 - ~\Apps\Node\node.EXE
    npm: 6.14.5 - ~\Apps\Node\npm.CMD
  npmPackages:
    clean-webpack-plugin: ^3.0.0 => 3.0.0 
    compression-webpack-plugin: ^4.0.1 => 4.0.1 
    html-webpack-plugin: ^4.3.0 => 4.3.0 
    optimize-css-assets-webpack-plugin: ^5.0.3 => 5.0.3 
    webpack: ^4.44.1 => 4.44.1 
    webpack-bundle-analyzer: ^3.8.0 => 3.8.0 
    webpack-cli: ^3.3.12 => 3.3.12 
    webpack-dev-server: ^3.11.0 => 3.11.0 
    webpack-merge: ^5.1.1 => 5.1.1 
</pre>

#### Debug info
How do you use this module? As CLI utility or as plugin?
-Both

If CLI, what command was used? (e.g. `webpack-bundle-analyzer -O path/to/stats.json`)

If plugin, what options were provided? (e.g. `new BundleAnalyzerPlugin({ analyzerMode: 'disabled', generateStatsFile: true })`)
[...]

PR #379, opened 2020-08-31
Add support to analyze gzipped bundles

xitter:
Feature request:  https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/377

New Config added:
```
  decompressExtenstion: {
    gz: { algorithm: 'unzipSync' },
    br: { algorithm: 'brotliDecompressSync' },
  },
```


README.md
```diff
@@ -68,6 +68,7 @@ new BundleAnalyzerPlugin(options?: object)
...
 |**`excludeAssets`**|`{null\|pattern\|pattern[]}` where `pattern` equals to `{String\|RegExp\|function}`|Default: `null`. Patterns that will be used to match against asset names to exclude them from the report. If pattern is a string it will be converted to RegExp via `new RegExp(str)`. If pattern is a function it should have the following signature `(assetName: string) => boolean` and should return `true` to *exclude* matching asset. If multiple patterns are provided asset should match at least one of them to be excluded. |
 |**`logLevel`**|One of: `info`, `warn`, `error`, `silent`|Default: `info`. Used to control how much details the plugin outputs.|
+|**`decompressExtenstion`**|`null` or `{Object}`|Default: `null`. Used to show stats for a compressed file, object has compressed file's extension as key and it's value is decompression algorithm of [zlib](https://nodejs.org/api/zlib.html) to use, like: `{ gz: { algorithm: 'unzipSync' } }`|
```
th0r, line 71: I don't think we need this option - decompression should be transparent to the end-user. It should just work without any configuration.

test/parseUtils.js
```diff
@@ -16,7 +21,7 @@ describe('parseBundle', function () {
     .forEach(bundleName => {
       it(`should parse ${_.lowerCase(bundleName)}`, function () {
         const bundleFile = `${BUNDLES_DIR}/${bundleName}.js`;
-        const bundle = parseBundle(bundleFile);
+        const bundle = parseBundle(bundleFile, {});
```
th0r, line 24: Options parameter should be optional.

test/parseUtils.js
```diff
@@ -26,8 +31,29 @@ describe('parseBundle', function () {
...
   });
+
+  Object.keys(COMPRESSIONS)
+    .forEach(compressionType => {
+      it(`should parse compressed ${compressionType} bundle`, function () {
+        const {extension, algorithm} = COMPRESSIONS[compressionType];
+        const bundleFile = `${BUNDLES_DIR}/validBundleWithArrowFunction.js`;
+        const compressedBundleFile = `${bundleFile}.${extension}`;
+        const expectedModules = JSON.parse(fs.readFileSync(`${BUNDLES_DIR}/validBundleWithArrowFunction.modules.json`));
+        if (!zlib[algorithm]) {
```
th0r, line 46: We don't need this condition - it swallows valid error and we won't notice it.

th0r (changes requested):

README.md
xitter: I too agree on this, but then should it basis extension only, i.e. gz and br?

test/parseUtils.js
xitter: For brotli, node version needs to be >= v11.7.0. Thus I put it to pass ci-checks which run on node 8. How do you suggest handling it?

th0r: > For brotli, node version needs to be >= v11.7.0
We need to mention about it in the docs and properly handle in the code.

th0r: > How do you suggest handling it?

xitter: > We need to mention it in the docs and properly handle in the code.

Brotli compression works for version >= v11.7.0,  so it is kind of an already set expectation for brotli assets. Would you prefer it including in docs anyway?

README.md
xitter: @th0r does the above comment look good?

test/parseUtils.js
th0r: Yes. Maybe compression is done with some external tool instead of Node.js.

PR #390, opened 2020-11-06
Replace `bfj` module with custom `stats.json` writer

th0r:
This PR replaces `bfj` module with custom `StatsSerializeStream` class which writes `stats.json` significantly faster (by about 6 times). Tested on 195mb `stats.json` file.

### DIFF
--- a/src/statsUtils.js
+++ b/src/statsUtils.js
@@ -32,7 +32,7 @@ class StatsSerializeStream extends Readable {
   * _stringify(obj) {
     if (typeof obj === 'string' || typeof obj === 'number' || typeof obj === 'boolean' || obj === null) {
       yield JSON.stringify(obj);
-    } else if (_.isArray(obj)) {
+    } else if (Array.isArray(obj)) {
       yield '[';
       this._indentLevel++;
 
--- a/src/statsUtils.js
+++ b/src/statsUtils.js
@@ -1,8 +1,6 @@
 const {createWriteStream} = require('fs');
 const {Readable} = require('stream');
 
-const _ = require('lodash');
-
 class StatsSerializeStream extends Readable {
   constructor(stats) {
     super();
@@ -49,7 +47,7 @@ class StatsSerializeStream extends Readable {
 
       this._indentLevel--;
       yield obj.length ? `\n${this._indent}]` : ']';
-    } else if (_.isPlainObject(obj)) {
+    } else {
       yield '{';
       this._indentLevel++;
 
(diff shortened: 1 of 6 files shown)


valscion:
Nice idea! :+1:

Would we want to assert on special object printing, too? Like `Symbol`, `Date`, regular expressions and functions?

Mostly that they won't be printed, like they won't be printed with `JSON.stringify`.

And other thing, would we want to also assert that stringifying the non-stringifiable things as being the only things would return `undefined` like `JSON.stringify` returns?

src/statsUtils.js
```diff
@@ -0,0 +1,84 @@
...
+      } else {
+        readMore = this.push(value);
+      }
+    }
+  }
+
+  * _stringify(obj) {
+    if (typeof obj === 'string' || typeof obj === 'number' || typeof obj === 'boolean' || obj === null) {
+      yield JSON.stringify(obj);
+    } else if (_.isArray(obj)) {
```
valscion, line 35: Do we need to use lodash here or would built-in `Array.isArray` work, too?

```suggestion
    } else if (Array.isArray(obj)) {
```

src/statsUtils.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        yield `${isFirst ? '' : ','}\n${this._indent}`;
+        yield* this._stringify(item);
+        isFirst = false;
+      }
+
+      this._indentLevel--;
+      yield obj.length ? `\n${this._indent}]` : ']';
+    } else if (_.isPlainObject(obj)) {
```
TrySound, line 52: There is the package which solves the problem problem
https://github.com/jonschlinkert/is-plain-object

th0r:
> Would we want to assert on special object printing, too? Like `Symbol`, `Date`, regular expressions and functions?
> 
> Mostly that they won't be printed, like they won't be printed with `JSON.stringify`.
> 
> And other thing, would we want to also assert that stringifying the non-stringifiable things as being the only things would return `undefined` like `JSON.stringify` returns?

`StatsSerializeStream` is meant to be used only to stringify webpack's stats object which is already pure JSON (it's a result of `stats.toJson()`). It means it can contain only primitives, arrays, plain objects or `null`s.

src/statsUtils.js
th0r: We use `lodash` in a lot of other places so there is no need to use other dependencies instead (at least for now).

src/statsUtils.js
th0r: Fixed

src/statsUtils.js
th0r: Actually, decided to remove this condition completely as it's not needed.

PR #393, opened 2020-11-08
Drop lodash from the package

TrySound:
Lodash is quite big package and often overused for things available
natively. In this diff I removed it from package dependencies and left
only for tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3036,6 +3036,23 @@
       "requires": {
         "each-props": "^1.3.0",
         "is-plain-object": "^2.0.1"
+      },
+      "dependencies": {
+        "is-plain-object": {
+          "version": "2.0.4",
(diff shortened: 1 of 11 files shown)


th0r:
Thanks for the PR, but I don't want to drop lodash as JS std library is not good enough to replace it yet and code becomes more complex.

TrySound:
Well, I spent a lot of time digging through not very obvious lodash solutions.

TrySound:
Also really in most cases it's just each -> forEach, filter -> filter, map -> map. Are these cases are less obvious?

TrySound:
Ok. Let's step away from code complication and I will do only drop in replacement.

onigoetz:
Hi, I'm a bit late to the party, is this PR still in progress ?

I'm also trying to remove lodash from my dependencies and would help reduce the install size of my tools.

PR #395, opened 2020-11-12
Avoid lodash where possible

TrySound:
In this diff I replaced some lodash methods with native ones without
complicating code.

(Related PR: #393)

### DIFF
--- a/src/analyzer.js
+++ b/src/analyzer.js
@@ -168,7 +168,7 @@ function readStatsFromFile(filename) {
 }
 
 function getChildAssetBundles(bundleStats, assetName) {
-  return bundleStats.children.find((c) =>
+  return (bundleStats.children || []).find((c) =>
     _(c.assetsByChunkName)
       .values()
       .flatten()
--- a/src/analyzer.js
+++ b/src/analyzer.js
@@ -48,7 +48,7 @@ function getViewerData(bundleStats, bundleDir, opts) {
   }
 
   // Picking only `*.js or *.mjs` assets from bundle that has non-empty `chunks` array
-  bundleStats.assets = _.filter(bundleStats.assets, asset => {
+  bundleStats.assets = bundleStats.assets.filter(asset => {
     // Filter out non 'asset' type asset if type is provided (Webpack 5 add a type to indicate asset types)
     if (asset.type && asset.type !== 'asset') {
       return false;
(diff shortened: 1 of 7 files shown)


TrySound:
@th0r Is this better?

valscion:
I would like to :+1: this idea of reducing the surface area of `lodash` in favor of native ones. I have had trouble deciphering what the lodash methods do as I'm myself more accustomed to the native methods.

src/analyzer.js
```diff
@@ -169,7 +168,7 @@ function readStatsFromFile(filename) {
 }
 
 function getChildAssetBundles(bundleStats, assetName) {
-  return _.find(bundleStats.children, (c) =>
+  return bundleStats.children.find((c) =>
```
th0r, line 171: Are you sure `bundleStats.children` is always defined?

src/parseUtils.js
```diff
@@ -342,9 +342,10 @@ function getModulesLocations(node) {
       node.arguments[0].elements :
       node.elements;
 
-    return _.transform(modulesNodes, (result, moduleNode, i) => {
-      if (!moduleNode) return;
+    return modulesNodes.reduce((result, moduleNode, i) => {
+      if (!moduleNode) return result;
```
th0r, line 346: Let's better inverse condition and return only once in the end of the function please.

src/utils.js
```diff
@@ -15,11 +15,11 @@ function createAssetsFilter(excludePatterns) {
         pattern = new RegExp(pattern, 'u');
       }
 
-      if (_.isRegExp(pattern)) {
+      if (types.isRegExp(pattern)) {
```
th0r, line 18: What is `types`?

th0r (changes requested):

src/utils.js
TrySound: New node.js utilities
https://nodejs.org/api/util.html#util_util_types

src/analyzer.js
TrySound: Nope, relied on tests. Will coalesce with []

src/utils.js
th0r: Nice, didn't know about it.

TrySound:
Hi @th0r is this good now?

th0r:
Yep, thanks!

PR #396, opened 2020-11-16
Upgrade gzip-size to version 6

realityking:
This removes the dependency on `pify` for `webpack-bundle-analyzer`.


th0r:
Thanks!

PR #397, opened 2020-11-24
Replace ejs templates with a simple js file

realityking:
This PR replaces the EJS based templates with a simple JS file using template strings. 

`webpack-bundle-analyzer` uses very little HTML templating since most of the client is a SPA. What is used, can be replaced by a straightforward JS file. This drops 15(!) packages from the dependencies. I also like how all functions related to generating the HTML get pulled into the same file.

### DIFF
--- a/src/template.js
+++ b/src/template.js
@@ -27,7 +27,7 @@ function getAssetContent(filename) {
 }
 
 function html(strings, ...values) {
-  return strings.map((string, index) => `${string}${values[index] || ''}`).join('')
+  return strings.map((string, index) => `${string}${values[index] || ''}`).join('');
 }
 
 function getScript(filename, mode) {
--- a/src/template.js
+++ b/src/template.js
@@ -26,6 +26,10 @@ function getAssetContent(filename) {
   return fs.readFileSync(assetPath, 'utf8');
 }
 
+function html(strings, ...values) {
+  return strings.map((string, index) => `${string}${values[index] || ''}`).join('')
+}
+
 function getScript(filename, mode) {
   if (mode === 'static') {
     return `<!-- ${_.escape(filename)} -->
@@ -36,7 +40,7 @@ function getScript(filename, mode) {
 }
 
 function renderViewer({title, enableWebSocket, chartData, defaultSizes, mode} = {}) {
-  return `<!DOCTYPE html>
+  return html`<!DOCTYPE html>
 <html>
   <head>
     <meta charset="UTF-8"/>
--- a/src/template.js
+++ b/src/template.js
@@ -1,22 +1,61 @@
-<!DOCTYPE html>
+/* eslint-disable max-len */
+const path = require('path');
+const fs = require('fs');
+
+const _ = require('lodash');
+
+const projectRoot = path.resolve(__dirname, '..');
+const assetsRoot = path.join(projectRoot, 'public');
+
+exports.renderViewer = renderViewer;
+
+/**
+ * Escapes `<` characters in JSON to safely use it in `<script>` tag.
+ */
+function escapeJson(json) {
+  return JSON.stringify(json).replace(/</gu, '\\u003c');
+}
+
+function getAssetContent(filename) {
+  const assetPath = path.join(assetsRoot, filename);
+
+  if (!assetPath.startsWith(assetsRoot)) {
+    throw new Error(`"${filename}" is outside of the assets root`);
+  }
+
+  return fs.readFileSync(assetPath, 'utf8');
+}
+
+function getScript(filename, mode) {
+  if (mode === 'static') {
+    return `<!-- ${_.escape(filename)} -->
+<script>${getAssetContent(filename)}</script>`;
+  } else {
+    return `<script src="${_.escape(filename)}"></script>`;
+  }
+}
+
+function renderViewer({title, enableWebSocket, chartData, defaultSizes, mode} = {}) {
+  return `<!DOCTYPE html>
 <html>
   <head>
     <meta charset="UTF-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1"/>
-    <title><%= title %></title>
+    <title>${_.escape(title)}</title>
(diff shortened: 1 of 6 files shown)


valscion:
Please fix lint errors ☺️

src/template.js
```diff
@@ -1,22 +1,61 @@
...
-      window.defaultSizes = <%- escapeJson(defaultSizes) %>;
+      window.chartData = ${escapeJson(chartData)};
+      window.defaultSizes = ${escapeJson(defaultSizes)};
     </script>
   </body>
-</html>
+</html>`;
+}
+
+exports.renderViewer = renderViewer;
```
th0r, line 61: Let's move it above `escapeJson` function declaration please so exported members become clearly visible.

src/template.js
```diff
@@ -1,22 +1,61 @@
...
+  if (mode === 'static') {
+    return `<!-- ${_.escape(filename)} -->
+<script>${getAssetContent(filename)}</script>`;
+  } else {
+    return `<script src="${_.escape(filename)}"></script>`;
+  }
+}
+
+function renderViewer({title, enableWebSocket, chartData, defaultSizes, mode} = {}) {
+  return `<!DOCTYPE html>
```
th0r, line 37: I see a few drawbacks of this approach:
- No syntax highlighting
- Much more difficult to do conditional branching e.g. `if/else` if we'll need it in the future.
- You have to remember to call `_.escape` manually

th0r (changes requested):

th0r:
> This drops 15(!) packages from the dependencies

What is the difference between `node_modules` sizes after `npm i --production`?

src/template.js
realityking: Done

th0r:
@realityking may I ask you to not use force-push in the future please as it significantly complicates tracking of changes.

src/template.js
th0r: I'm not saying that I'm strongly against this PR, no. We can live with these drawbacks, but only if it brings some significant improvements. That's why I asked to provide difference in `node_modules` sizes.

realityking: - Syntax highlighting could be added in some editors/plugins by adding a "fake" tag `html`. I'll add that as a commit so you can see what it looks like.
- Agreed that conditionals are more cumbersome but do you expect the templates to get significantly more complex than they are today?
- This is true (though you also have to be mindful of `<%=` vs `<%-` in EJS) and I wouldn't do it for a large server rendered app but as of today there's just two variables in the templates here that need to be escaped.

To me, this is trade-off of complexity. EJS is certainly nicer than template literals but for a template this small it's, IMHO, not worth the cost of the dependencies. Some of which are very old and don't even deduplicate.

TrySound:
ejs started to depend on jake build took since v3 for some reason. v2 was dependency free.

realityking:
Sorry about the force push, I'll refrain from them.

The size difference in `node_modules` is 1MB with this patch. (11 MB > 10 MB)

realityking:
Just to further illustrate this point, these are the dependencies that would no longer be required:

```
├─┬ ejs@3.1.5
│ └─┬ jake@10.8.2
│   ├── async@0.9.2
│   ├─┬ chalk@2.4.2
│   │ ├─┬ ansi-styles@3.2.1
│   │ │ └─┬ color-convert@1.9.3
│   │ │   └── color-name@1.1.3
│   │ ├── escape-string-regexp@1.0.5
│   │ └─┬ supports-color@5.5.0
│   │   └── has-flag@3.0.0
│   ├─┬ filelist@1.0.1
│   │ └── minimatch@3.0.4 deduped
│   └─┬ minimatch@3.0.4
│     └─┬ brace-expansion@1.1.11
│       ├── balanced-match@1.0.0
│       └── concat-map@0.0.1
```

th0r:
> ejs started to depend on jake build

Hmm, I wonder why do they need build tool as runtime dependency?

TrySound:
They use it to build cli https://github.com/mde/ejs/commit/82a030985891ad7efcfb6be381aeb742c1d25c81#diff-ca900686fca4680d4692bf587dc5da1322879c344688c187b5511cda27902c18R20

src/viewer.js
```diff
@@ -140,42 +133,28 @@ async function generateReport(bundleStats, opts) {
   if (!chartData) return;
 
   await new Promise((resolve, reject) => {
-    ejs.renderFile(
-      `${projectRoot}/views/viewer.ejs`,
-      {
+    try {
```
TrySound, line 136: Looks like there is no need in promise here anymore. The code is sync

valscion:
This approach is looking really promising to me! I'd be happy if we could make this pipeline easier by removing the ejs templates.

I'm thrilled to see your contributions, @TrySound ❤️. I hope you're not feeling discouraged as we are a bit slow to look through your changes — rest assured that we are happy to see you contribute to webpack-bundle-analyzer 😊

realityking:
@valscion I was confused when I got the email 😄

So far, I'm really not concerned about review times. I realize nobody is working full-time on this. 🙂

src/viewer.js
th0r: It will be a breaking change.

TrySound: I mean the function can be still async but promise wrapper around sync code is not necessary.

th0r (approved):

th0r:
Ok, let's merge this. I think removing an extra dependency outweighs the drawbacks. @valscion WDYT?

src/viewer.js
th0r: Ah, yep, agree. @realityking could you change it please?

valscion (approved):
My thoughts: :tada: :smile:

src/viewer.js
realityking: Pushed a commit to do this :)

src/viewer.js
```diff
@@ -139,43 +132,23 @@ async function generateReport(bundleStats, opts) {
...
+
+  fs.mkdirSync(path.dirname(reportFilepath), {recursive: true});
+  fs.writeFileSync(reportFilepath, reportHtml);
+
+  logger.info(`${bold('Webpack Bundle Analyzer')} saved report to ${bold(reportFilepath)}`);
+
+  if (openBrowser) {
+    open(`file://${reportFilepath}`, logger);
+  }
   });
```
TrySound, line 152: this line should not be here I think

realityking: You're right. I wonder how the tests worked with this line present 🤔

TrySound: They didn't. Travis hangs.

th0r (approved):

valscion (approved):
Yeah this looks good to me! If you feel like adding a changelog entry, that would be amazing ☺️

realityking:
Done :)

valscion:
I'll close and reopen the PR to see if Travis build would start up... I want to merge this PR :smile:

realityking:
@valscion It looks like `webpack-bundle-analyzer` is still on travisci.***org***. That will stop working entirely on December 31st but it might already cause performance issues: https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#q-why-are-some-of-my-queued-jobs-taking-longer-than-usual-to-build

Switching to ***.com***, or somewhere else entirely, would be a good idea.

alexander-akait:
Maybe github actions is better idea, most of webpack repos already migrate on it

valscion:
> Maybe github actions is better idea, most of webpack repos already migrate on it

Yeah I have been wondering if github actions would be a good idea. A PR for that would be appreciated ☺️

valscion:
I merged this PR now as I got tired of waiting on Travis.

realityking:
Thanks @valscion :)

Let's hope I didn't overlook anything. Please ping in case master does error out and I'll fix it ASAP.

valscion:
Released in v4.2.0 :relaxed:. Thank you for your contributions!

PR #398, opened 2020-11-28
Replace express with builtin node server

TrySound:
Alternative to long-standing https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/212

Express in this project does not add much value. Middleware api is not
necessary for 1.5 middlwares. We can try to use node server directly.

Serving static is left for express middleware. Can be replaced with
something smaller in another PR.

Thanks to @realityking for inspiration


src/viewer.js
```diff
@@ -13,6 +12,7 @@ const {open} = require('./utils');
 const {renderViewer} = require('./template');
 
 const projectRoot = path.resolve(__dirname, '..');
+const viewsRoot = path.join(projectRoot, 'views');
```
realityking, line 15: This line is not needed (merge artefact)

package.json
```diff
@@ -37,11 +37,11 @@
     "acorn-walk": "^8.0.0",
     "chalk": "^4.1.0",
     "commander": "^6.2.0",
-    "express": "^4.17.1",
     "filesize": "^6.1.0",
     "gzip-size": "^6.0.0",
     "lodash": "^4.17.20",
     "opener": "^1.5.2",
+    "serve-static": "^1.14.1",
```
valscion, line 44: Seems like as per this comment:

https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/212#discussion_r449811403

`serve-static` is a big reason why `express` is so large so if polka would work for our use case, it's much smaller than using `serve-static`.

valscion: Oh, that other PR uses `sirv` for the static serving?

TrySound: Well, if you are ok to introduce sirv I will use it. I just remember there was a doubt to introduce polka packages.

valscion: Oh, I don't really know what opinion @th0r has 😅 — sorry for being unclear. If things work then the size differences this provides might be useful?

As long as the server is still able to show new results when a bundle is changed like was the purpose behind introducing this feature back in https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/74 then I would be OK with this change.

I think the supported node versions are good enough, too, and we don't accidentally regress from supporting node v10.13.0

TrySound: To support watch I passed "dev" flag which prevents caching.
I guess the size will be almost the same but without more polka dependencies.
https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/212#issuecomment-735806643

valscion: The `serve-static` is still in dependencies so I don't know how to figure out the size difference yet — some numbers on how this potentially affects the size of `node_modules` would be nice.

TrySound: https://packagephobia.com/result?p=static-serve
https://packagephobia.com/result?p=express (depends on static-serve out of the box)
https://packagephobia.com/result?p=sirv
https://packagephobia.com/result?p=polka (does not include sirv out of the box)
https://packagephobia.com/result?p=%40polka%2Furl (dependency of both sirv and polka)

src/viewer.js
```diff
@@ -48,22 +48,34 @@ async function startServer(bundleStats, opts) {
...
+      res.writeHead(200, {'Content-Type': 'text/html'});
+      res.end(html);
+    } else {
+      sirvMiddleware(req, res, err => {
+        if (err) {
+          console.error(err.stack || err.toString());
+          res.writeHead(500);
+          res.end();
+        } else {
+          res.writeHead(404);
```
th0r, line 73: Why 404?

src/viewer.js
```diff
@@ -48,22 +48,34 @@ async function startServer(bundleStats, opts) {
...
-    const html = renderViewer({
-      mode: 'server',
-      title: resolveTitle(reportTitle),
-      chartData,
-      defaultSizes,
-      enableWebSocket: true
-    });
-    return res.end(html);
+  const sirvMiddleware = sirv(`${projectRoot}/public`, {
+    dev: true
```
th0r, line 52: What does this flag mean? Let's add a comment about it please.

th0r (changes requested):

src/viewer.js
TrySound: Yeah, callback is not necessary. This is default behaviour. Removed.

src/viewer.js
TrySound: It prevents caching for watch mode. Added comment.

th0r (approved):

valscion (approved):

PR #401, opened 2020-12-01
Move filesize to dev dependencies

realityking:
It's only used in the client


TrySound:
Is client bundled before publishing?

package.json
```diff
@@ -37,7 +37,6 @@
     "acorn-walk": "^8.0.0",
     "chalk": "^4.1.0",
```
alexander-akait, line 38: Can be replaced by `colorette` (light, fast and with simple API), already used in `webpack-cli@4`, so for developers which use `webpack-cli` and `webpack-bundle-analyzer` we will have less deps

realityking:
@TrySound Yes: https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/package.json#L23

package.json
realityking: Good point but something for a different PR :)

valscion (approved):
Cool :+1: Just checked that `mobx` is also only in devDependencies so client-only packages are OK to be specified only in devDependencies :+1:

Issue #400, opened 2020-11-30
Help us migrate CI to GitHub Actions


valscion:
Basically, we're using travis-ci.org and it is getting sunset in December 31st: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/397#issuecomment-735797582

We probably want to move the CI to GitHub actions. Most of webpack repos have already migrated to it: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/397#issuecomment-735798563

PR #402, opened 2020-12-02
Replace Travis with GitHub actions as CI

valscion:
Resolves https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/400


valscion:
I disabled the `continuous-integration/travis-ci` branch protection for now so that this PR is not blocked on advancing. If @th0r you feel like this is good to go, I can add branch protections for the GitHub actions if necessary.

th0r:
I wonder is it possible to add `Node.js` prefix to version numbers as now it's a bit not obvious what do these numbers mean?

image

valscion:
I don't know — I can check if that's possible when I have the time ☺️

valscion:
It now looks like so:

Replace_Travis_with_GitHub_actions_as_CI_by_valscion_·_Pull_Request__402_·_webpack-contrib_webpack-bundle-analyzer

It seems quite simple to change the name to whatever we want as can be seen by fdb5a0e08b102ca1e1519d2995a35a3b84c4758b — feel free to make a direct code suggestion to this PR if you have a better wording in mind ☺️

.github/workflows/main.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  build-and-test:
+    strategy:
+      matrix:
+        node:
+          - '10'
+          - '12'
+          - '14'
+    runs-on: ubuntu-latest
+    name: Tests on node ${{ matrix.node }}
```
th0r, line 16: Just a small suggestion:
```suggestion
    name: Tests on Node.js v${{ matrix.node }}
```

th0r (changes requested):

th0r (approved):
Great, thanks!

valscion:
> I can add branch protections for the GitHub actions if necessary.

Do we need the branch protections? I'm not sure if they're necessary — a red CI should be enough for us to avoid merging pull requests.

The CLA-check of course stays as a required status check.

th0r:
> Do we need the branch protections?

I don't even know what is that =)

valscion:
Ah, it's this one:

feat__Render_Selected_Module_Size_Total_by_vwb_·_Pull_Request__346_·_webpack-contrib_webpack-bundle-analyzer

So one can't merge a pull request (unless they're an admin and admin overrule is allowed) if a required status check is not successful.

We could configure some or all of the GitHub Actions status checks to be required to be successful to allow merging via https://github.com/webpack-contrib/webpack-bundle-analyzer/settings/branches, from here:

Branch_protection_rule

th0r:
Ah, ok, thanks for the explanation. I don't think we need it.

valscion:
Yeah I think the same, let's keep it as it is now :+1:

PR #413, opened 2021-01-19
Hide tooltip when hovering over the attribution group and between top level groups.

stanislawosinski:


valscion:
Thanks! Would you be able to showcase the difference in behavior with a small screencap or something? It would make it easier to understand how this behavior has changed now ☺️

valscion:
Thanks! Would you be able to showcase the difference in behavior with a small screencap or something? It would make it easier to understand how this behavior has changed now ☺️

stanislawosinski:
Before:

https://user-images.githubusercontent.com/628360/105147827-fb0d9b80-5b01-11eb-9c6a-f0241c86ed34.mp4

After:

https://user-images.githubusercontent.com/628360/105147856-0365d680-5b02-11eb-9713-0aa3e54a0b40.mp4

valscion:
Nice, thank you! Looks great!

Would you be able to write a small note on the changelog about this?

client/components/Treemap.jsx
```diff
@@ -105,9 +105,12 @@ export default class Treemap extends Component {
       },
       onGroupDoubleClick: preventDefault,
       onGroupHover(event) {
-        // Ignoring hovering on `FoamTree` branding group
-        if (event.group && event.group.attribution) {
+        // Ignoring hovering on `FoamTree` branding group and the root group
+        if (event.group && (event.group.attribution || event.group === this.get('dataObject'))) {
           event.preventDefault();
+          if (props.onMouseLeave) {
```
valscion, line 111: Do we need this conditional? Is there a place in the code where `props.onMouseLeave` would not be specified?

stanislawosinski: I made the condition analogous to the `props.onGroupHover` check below, which is also there but probably not required:

```
        if (props.onGroupHover) {
          props.onGroupHover.call(component, event);
        }
```

I can remove the `props.onMouseLeave` condition, not a problem.

th0r (approved):

th0r:
Thanks!

PR #414, opened 2021-01-19
Draw FoamTree labels also during animation

stanislawosinski:
This should improve the user experience and should still be fairly smooth on recent hardware.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,10 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 ## UNRELEASED
 
-<!-- Add changelog entries for new changes under this section -->
-
+* **Improvement**
+  * Keep treemap labels visible during zooming animations for better user experience ([#414](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/414) by [@
+    stanislawosinski](https://github.com/stanislawosinski))
+    
 ## 4.3.0
 
  * **Improvement**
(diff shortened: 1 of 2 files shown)


valscion:
Thanks! Would you be able to showcase the difference in behavior with a small screencap or something? It would make it easier to understand how this behavior has changed now ☺️

stanislawosinski:
Before:

https://user-images.githubusercontent.com/628360/105148129-50e24380-5b02-11eb-82f8-f660ddcd1f1b.mp4

After:

https://user-images.githubusercontent.com/628360/105148174-5b9cd880-5b02-11eb-85d2-c0d618016292.mp4

valscion:
Ooh neat! Thanks for the video, this helps me understand this change a lot better ☺️ 

Would you be able to write a small changelog entry about this change?

th0r:
I'm not sure about this change as it may make a significant negative impact on the rendering performance during the animation for the large bundles especially on high-dpi screens e.g. retina.

stanislawosinski:
Actually, the zooming animation is rendered with pixelRatio=1. On High-DPI screens you should see a faded switch between the low- and high-resolution rendering before and after the animation. 

Additionally, when the number of groups is large, FoamTree does not render labels if their on-screen size is smaller than a threshold. The labels render only when the label size increases as a result of zooming.

I was thinking of adding a configuration checkbox that would enable/disable labels during zooming, but then decided the gain may not be worth cluttering the UI. Ultimately, it's a matter of testing the performance on typical developer configurations. The no-labels zoom is a default that dates back to the first versions of FoamTree released nearly 10 years ago. The hardware has progressed a lot since then, so it may be a good moment to rethink this behaviour :-)

th0r:
Ok, I've checked that and looks like the rendering slowdown is acceptable. Could you fix a conflict in changelog please?

th0r (approved):

th0r:
Thanks!

PR #415, opened 2021-01-26
feat: reuse the same tab on Chrome

xiaoxiaocoder:


th0r:
Thanks for the PR, but I'm actually relying on the current "new tab" logic quite often - I leave tabs with previous reports opened to compare them with the new report. So I would vote against this feature, sorry.

xiaoxiaocoder:
Ok, got it. ;)

close #415

xiaoxiaocoder:
> Thanks for the PR, but I'm actually relying on the current "new tab" logic quite often - I leave tabs with previous reports opened to compare them with the new report. So I would vote against this feature, sorry.

ok, got it ;)

PR #423, opened 2021-03-19
Support reading large stats.json files

henry-alakazhang:
Node has a string size limit of 500MB and the current `JSON.parse(fs.readFileSync(...))` logic to read `stats.json` files fails for files >500MB.

This PR changes the approach for stats parsing to use a stream, similar to this PR for webpack-cli in creating the file: https://github.com/webpack/webpack-cli/pull/2190

I've tested it against a local repo with a 700MB stats.json and it works when it didn't before. I would add a test, but I don't know about adding that large of a file :joy: Could look into generating one if you would like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 ## UNRELEASED
 
+ * **Improvement**
+   * Support reading large (>500MB) stats.json files ([#423](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/423) by [@henry-alakazhang](https://github.com/henry-alakazhang))
+
 ## 4.7.0
 
  * **New Feature**
--- a/package-lock.json
+++ b/package-lock.json
@@ -1258,6 +1258,12 @@
       "integrity": "sha512-cDql5T3eS4Y5jzrDj25QNpU0pHgLFJ96MaoMAkJX8ceL2yat4u7EbIuqLoTbQLiTy1KxDRFbxHIKLzTUT64Duw==",
       "dev": true
     },
+    "@discoveryjs/json-ext": {
+      "version": "0.5.7",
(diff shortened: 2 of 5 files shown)


cblnco:
Is there a possibility this feature this gets merged @th0r, @valscion?
I know that the workaround would be to use the plugin instead of the CLI. However, I'm on a scenario where my project doesn't manages all the Webpack settings directly and I can't use the plugin. 
So It would be really really nice to use the CLI too.

valscion:
Yeah this looks like something that could work! I seem to have forgotten that this PR exists, sorry about that 😳 

@henry-alakazhang would you be open to:

* recreating this pull request on top of the current `master` branch
* use the latest version of `@discoveryjs/json-ext` that exists
* create a changelog entry as "Improvement" category
* signing the CLA

We can make do without a test for this as long as the existing tests pass.

gavinbarron:
@valscion this seems to be stalled, and it's something that we could make use of for tracking work to produce smaller app bundles when using [kiota](https://github.com/microsoft/kiota) to build TypeScript SDKs for HTTP APIs.

If @henry-alakazhang is not able to progress this contribution I'd be happy to pick it up and move it forward.

cblnco:
Now that I have bandwidth I could also help moving forward with this contribution @valscion, @henry-alakazhang, @gavinbarron.
I will keep an eye on this to know who ends up integrating these changes.

henry-alakazhang:
Hey whoops sorry, completely missed this. I'll update the PR.

henry-alakazhang:
Updated with latest version of `json-ext`. I've refactored the code to use async/await because I wasn't sure how `readStatsFromFile` would actually throw errors.

Remains untested as I can't run `yarn test` for some reason - the `npm i` to install the test webpack versions doesn't want to run regardless of which Node version I'm using (erroring with `Maximum call stack size exceeded`).

valscion (approved):
Thanks, this looks great and the tests still pass in CI so I'll merge :)

valscion:
This is now in v4.8.0! 🚀

PR #426, opened 2021-03-24
docs: udpate webpack command

snitin315:
update webpack command for generating `stats.json`, https://webpack.js.org/api/cli/#json


valscion:
This is not compatible with webpack 4: https://v4.webpack.js.org/api/cli/#common-options

snitin315:
Ah yes we don't `--profile` for webpack 4 with webpack-cli v4.

PR #428, opened 2021-03-25
chore: quote star in issue template command

cncolder:
it throws `no matches found: --npmPackages=*webpack*` without this


valscion:
Thanks!

PR #432, opened 2021-04-10
feat: Brotli compression support

dcsaszar:
This is my go at #113

~~Instead of _adding_ Brotli support, I added low-level options to _replace_ Gzip support. The overall approach is inspired by me trying to avoid the hassle of conditional code based on the Node.js version, my own requirements (no need for high-level options), and https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/113#issuecomment-611766511:~~
> I have the use-case that the pipeline analysis doesn't use gzip at all, it relies completely on brotli

Updated with high-level option `compressionAlgorithm: gzip | brotli`

### Example usage:

```js
new BundleAnalyzerPlugin({
  defaultSizes: "compressed",
  compressionAlgorithm: "brotli",
})
```

### Regarding the requirements...
...stated in https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/113#issuecomment-479513104

> The only concern I have is the increasing of analyze time. Agree with @valscion here - it would be nice to see a proof-of-concept PR and comparison of analyze times with and without it.

For me, the POC was trying out https://www.npmjs.com/package/webpack-bundle-analyzer-brotli, which works great, but is outdated.
The new option is opt-in. Either gzip or Brotli. Users who want Brotli support will be willing to accept the longer wait.

> It might be possible to add brotli support if it would be calculated only when some CLI or plugin option would be there. That way it would not increase analyze time for those who don't yet care about Brotli.

The user must specify the compression as an option. They know their environment (e.g. the Node.js version) and resources best.

> The crux is that we don't want this to increase analyze time, and this feature would have to be opt-in. Currently it might require some architectural changes, though, so it isn't that easy to do.

I tried to keep the changes at a minimum.
In this PR I even left most of the internals (including a lot of `gzip*` naming) untouched. A cleanup of the internal naming could be done independently, but I'm not sure if this is required.


valscion:
Thanks for taking a stab at this!

I'm checking out the PR description and the API changes now, and are not focusing on the internal code changes yet as I write this comment, so consider this comment a high-level review ☺️ 

I'm a bit concerned how this API separetes the `compressedSize` and `compressedSizeLabel` options from each other. I'd hope we could have an option like `compressionType` or `compressionAlgorithm` instead where one could choose `gzip` or `brotli` for now, and it would then automatically set the label.

I can see that we could then later, after this pull request, consider opening up the API more and allow some sort of a `compressionAlgorithm: {function}` API — and then we'd have to figure out how we want to handle the label in that case.

For now, to ease reviewing, I'd favor a less open API change where we could focus only on choosing between `gzip` and `brotli`. We could later discuss the "custom algorithm" option, if we want that.

This "brotli" or "gzip" choice could very well be implemented only at the CLI and Plugin options levels — the internal wiring could keep on using the `compressedSize` and `compressedSizeLabel` options you've implemented here. So my comment about only allowing the choice between gzip and brotli would affect only the code in CLI and Plugin options handling, and not the internal wiring.

What do you think? Am I making any sense? 😅 

---

Oh, and as you can see, the tests have become unfortunately quite flaky indeed. I haven't had the time to look into why the tests are timing out or getting aborted. Any help in getting the GitHub actions to run properly would be very much appreciated and would also help reviewing this PR once we're happy with the way the feature has been designed :pray:

We haven't really done anything about the CI since https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/402 where the checks were running successfully. So something has likely changed since then, and it likely isn't anything we've written as there isn't quite much git history since removing Travis CI: https://github.com/webpack-contrib/webpack-bundle-analyzer/commits/master

dcsaszar:
`compressionAlgorithm: 'gzip' | 'brotli'` sounds good to me. I'll give it a go.

valscion:
The CI is now hopefully fixed, thanks to #435. Feel free to merge or rebase newest changes in to this PR to hopefully get a green CI ☺️

dcsaszar:
@valscion ready for the next review!

src/sizeUtils.js
```diff
@@ -0,0 +1,13 @@
+const zlib = require('zlib');
+
+export function gzipSize(input) {
+  return zlib.gzipSync(input, {level: 9}).length;
+}
+
+export function brotliSize(input) {
+  if (typeof zlib.brotliCompressSync !== 'function') {
+    throw new Error('Brotli compression requires Node.js v10.16.0 or higher.');
+  }
```
valscion, line 10: Nice :+1:

test/plugin.js
```diff
@@ -170,6 +170,45 @@ describe('Plugin', function () {
...
+      it('should support brotli', async function () {
+        const config = makeWebpackConfig({
+          analyzerOpts: {
+            compressionAlgorithm: 'brotli'
+          }
+        });
+        await webpackCompile(config, '4.44.2');
+        await expectValidReport({
+          parsedSize: 1311,
+          gzipSize: 302
```
valscion, line 208: I think that, even though it increases the size of this PR, it might be better to aim to use `brotliSize` in the data when Brotli is used instead of `gzipSize`. It would be quite confusing for direct users of the report JSON file (e.g. those who supply `--analyzerMode json` value to CLI) if the key would remain as `gzipSize` even though Brotli was selected.

What do you think? Does the confusion risk outweigh changing the code more?

src/viewer.js
```diff
@@ -126,21 +136,23 @@ async function generateReport(bundleStats, opts) {
     openBrowser = true,
     reportFilename,
     reportTitle,
+    compressionAlgorithm = 'gzip',
```
valscion, line 139: Oh yeah, I see you've found that we have multiple places where we're defining default values for options, and haven't really made up our mind where they'd belong 😅 

How do you feel about moving all the `compressionAlgorithm = 'gzip` default value definitions to the top-level usages? That is, moving the `compressionAlgorithm = 'gzip'` code to the plugin starting point in `src/BundleAnalyzerPlugin.js`. The CLI starting point in `src/bin/analyzer.js` already seems to define the default value for the `--compression-algorithm` to be `gzip` so we should be covered there.

Then we could get rid of all the `compressionAlgorithm = 'gzip'` defaulting in other code paths and assume that we have the value filled in in those code paths.

README.md
```diff
@@ -147,9 +150,9 @@ It is called "stat size" because it's obtained from Webpack's
 This is the "output" size of your files. If you're using a Webpack plugin such
 as Uglify, then this value will reflect the minified size of your code.
 
-### `gzip`
+### `compressed`
```
valscion, line 153: Can we make it so that the added option for `defaultSizes` would be `brotli` instead of `compressed`? That way the current, default gzip usage would not have to be changed.

That way we could keep the gzip-related documentation intact and only worry about adding new documentation related to the brotli options.

I could imagine that the fourth size definition documentation here could look something like this:

> ### `brotli`
> 
> This is the size of running the parsed bundles/modules through [Brotli](https://www.brotli.org/) compression. **Note:** This needs the compression algorithm to be configured to use Brotli.

valscion (commented):
I like the direction of this PR :+1:. It was a good idea to get rid of the `gzip-size` package and rely on the native `zlib` module also for the gzip case — it makes the code paths for brotli and gzip look a bit more similar, and all the previous usages of `gzip-size` were easy to find :+1:

test/plugin.js
th0r: Maybe just `compressedSize`?

client/components/ModulesTreemap.jsx
```diff
@@ -18,11 +18,13 @@ import Search from './Search';
...
-const SIZE_SWITCH_ITEMS = [
-  {label: 'Stat', prop: 'statSize'},
-  {label: 'Parsed', prop: 'parsedSize'},
-  {label: 'Gzipped', prop: 'gzipSize'}
-];
+function allSizeSwitchItems() {
+  return [
+    {label: 'Stat', prop: 'statSize'},
+    {label: 'Parsed', prop: 'parsedSize'},
+    {label: window.compressedSizeLabel, prop: 'gzipSize'}
```
th0r, line 25: Can we change `gzipSize` to `compressedSize`?

src/analyzer.js
```diff
@@ -102,7 +111,7 @@ function getViewerData(bundleStats, bundleDir, opts) {
 
     if (assetSources) {
       asset.parsedSize = Buffer.byteLength(assetSources.src);
-      asset.gzipSize = gzipSize.sync(assetSources.src);
+      asset.gzipSize = compressedSize(assetSources.src);
```
th0r, line 114: ?
```suggestion
      asset.compressedSize = compressedSize(assetSources.src);
```

src/analyzer.js
```diff
@@ -20,11 +25,15 @@ module.exports = {
...
   const {
     logger = new Logger(),
-    excludeAssets = null
+    excludeAssets = null,
+    compressionAlgorithm
   } = opts || {};
 
   const isAssetIncluded = createAssetsFilter(excludeAssets);
 
+  const compressedSize = COMPRESSED_SIZE[compressionAlgorithm];
```
th0r, line 34: Let's better call it `getCompressedSize` please

src/analyzer.js
```diff
@@ -143,7 +152,7 @@ function getViewerData(bundleStats, bundleDir, opts) {
     }
 
     asset.modules = assetModules;
-    asset.tree = createModulesTree(asset.modules);
+    asset.tree = createModulesTree(asset.modules, {compressedSize});
```
th0r, line 155:
```suggestion
    asset.tree = createModulesTree(asset.modules, {getCompressedSize});
```

src/viewer.js
```diff
@@ -22,6 +22,14 @@ function resolveTitle(reportTitle) {
   }
 }
 
+function resolveDefaultSizes(defaultSizes) {
+  return defaultSizes === 'compressed' ? 'gzip' : defaultSizes;
```
th0r, line 26: Can we inverse it please? I mean, deprecate `gzip` default size and use `compressed` instead to it will look like this:
```suggestion
  return defaultSizes === 'gzip' ? 'compressed' : defaultSizes;
```

th0r (changes requested):

src/viewer.js
dcsaszar: My current approach, based on https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/432#discussion_r613798534, is:
```js
function resolveDefaultSizes(defaultSizes, compressionAlgorithm) {
  if (['gzip', 'brotli'].includes(defaultSizes)) return compressionAlgorithm;
  return defaultSizes;
}
```

That is, we drop `compressed`. The config is "forgiving" if the user specifies the wrong `defaultSizes`.
WDYT?

test/plugin.js
dcsaszar: @th0r I thought about this too, as I'm not a friend of key explosion.

On the other hand I'm OK with `gzipSize`, `brotliSize`:
* backwards compatible
* theoretically allows for dual-compression stats

dcsaszar:
@valscion @th0r next round.

README.md
```diff
@@ -61,7 +61,8 @@ new BundleAnalyzerPlugin(options?: object)
 |**`analyzerPort`**|`{Number}` or `auto`|Default: `8888`. Port that will be used in `server` mode to start HTTP server.|
 |**`reportFilename`**|`{String}`|Default: `report.html`. Path to bundle report file that will be generated in `static` mode. It can be either an absolute path or a path relative to a bundle output directory (which is output.path in webpack config).|
 |**`reportTitle`**|`{String\|function}`|Default: function that returns pretty printed current date and time. Content of the HTML `title` element; or a function of the form `() => string` that provides the content.|
-|**`defaultSizes`**|One of: `stat`, `parsed`, `gzip`|Default: `parsed`. Module sizes to show in report by default. [Size definitions](#size-definitions) section describes what these values mean.|
+|**`defaultSizes`**|One of: `stat`, `parsed`, `gzip`, `brotli`|Default: `parsed`. Module sizes to show in report by default. [Size definitions](#size-definitions) section describes what these values mean.|
```
th0r, line 64: In this case user will have to provide compression algorithm twice: for `compressionAlgorithm` and `defaultSizes` and it opens possibilities for invalid option combinations like `defaultSizes: 'gzip', compressionAlgorithm: 'brotli'` which don't make sense. I personally would change `defaultSizes: 'gzip'` to `defaultSizes: 'compressed'` and deprecate the former, but support it (convert to `compressed`).

client/components/ModulesTreemap.jsx
```diff
@@ -18,11 +18,17 @@ import Search from './Search';
...
-  {label: 'Parsed', prop: 'parsedSize'},
-  {label: 'Gzipped', prop: 'gzipSize'}
-];
+function allSizeSwitchItems() {
+  const items = [
+    {label: 'Stat', prop: 'statSize'},
+    {label: 'Parsed', prop: 'parsedSize'}
+  ];
+
+  if (window.compressionAlgorithm === 'gzip') items.push({label: 'Gzipped', prop: 'gzipSize'});
```
th0r, line 27: I would change `gzipSize/brotliSize ` to `compressedSize` but looks like it will be a breaking change.

th0r (changes requested):

README.md
dcsaszar: The current implementation handles the mismatch gracefully, so it's more of a cosmetic issue on the user side.

Reintroducing `compressed` would help here. If used, the config always looks "good", even when switching the algorithm. _If_ we support `compressed`, fail on mismatches between `compressionAlgorithm` and `defaultSizes`?

valscion:
I'll let you @th0r review this ☺️ — I think this is heading into a good direction and trust your judgement here :+1:

client/components/ModulesTreemap.jsx
valscion: Would be nice if we could first do this as a new feature, release in a minor version and then do a major release later to consolidate the API on on plain `compressedSize`. That has been my opinion during this PR review at least ☺️.

What do you think of this plan?

RodrigoTomeES:
Hi, Is there any update on this functionality? Thanks for your work!

pumano:
Amazing feature not going to production just by personal opinions on how to name it brotli or general word "compression".

dcsaszar:
@valscion in https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/432#issuecomment-820386722 you handed the review to @th0r. 
I see his last PR was #449 so I'm not sure he's still active/in charge. I'd like to continue working on this, but the responsibilities and todos should be clear.

valscion:
Yeah looks like I can take over reviewing this. Now that brotli is used more and more, having this show in `webpack-bundle-analyzer` would definitely be useful.

Let's first get this PR up-to-date and then I can start reviewing again. I've lost context of this PR so I'll basically have to start the review process from scratch as well

RuwaidLouisHejare:
Fantastic to see this feature still is active. I would love to see this in production. Keep up the great work!

lukemwillis:
Any updates on this feature getting merged? Would love to see brotli compression in bundle analyzer.

dcsaszar:
Closing in favor of #663

@lukemwillis Thanks for the nudge!

PR #433, opened 2021-04-11
Fixes #417 missing module chunks

### LINKED ISSUE #417: Stats parsing is broken when using "all: true" and Webpack 5
Hello!

Thank you for this great tool!

#### Issue description

However, I've encountered a situation when stats file parsing is broken when using the following stats configuration:

```javascript
const statsConfig = {
  all: true,
  context: projectPath,
  source: false,
}
```

and Webpack `5.22.0`.

#### Technical info

The problem is in the [assetHasModule](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/37f048a500ebaa2394350183dee7b5c4c09d8489/src/analyzer.js#L191) function:

```javascript
function assetHasModule(statAsset, statModule) {
  // Checking if this module is the part of asset chunks
  return statModule.chunks.some(moduleChunk => statAsset.chunks.includes(moduleChunk));
}
```

It raises the following error:

`TypeError: Cannot read property 'some' of undefined`

because `statModule.chunks` is `undefined` in this case.

In my example the module stat object has the following structure:

```json5
{
  "type": "modules by path",
  "name": "…node_modules/.pnpm/core-js@3.8.3/node_modules/core-js/internals/*.js",
  "children": [
    // A big list of modules here
  ],
  "size": 22868,
  "sizes": {
    "javascript": 22868
  }
}
```

---

Also, when I change the stats config to use "verbose" preset, the error goes away, but when I view the bundle layout in "parsed" format, it only renders the entry point module. I think something is broken on a more fundamental level.

---

[I've uploaded](https://drive.google.com/file/d/1wpWsx6ELqt0Dy5kfHnlylyOnd3QQtD6Z/view) the obfuscated version of my stats.json file generated with the latest version of the Webpack using "verbose" preset.

deanshub:
Fixes #417 missing module chunks


valscion:
Hello, and thank you for your contribution!

The code does look good to me. However, I'm a bit worried about our CI failing, likely not due to something you've done but something that is out of our control right now.

I'm copy-pasting from my comment in another PR (this one: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/432#issuecomment-817570076):

> I haven't had the time to look into why the tests are timing out or getting aborted. Any help in getting the GitHub actions to run properly would be very much appreciated and would also help reviewing this PR once we're happy with the way the feature has been designed 🙏
> 
> We haven't really done anything about the CI since #402 where the checks were running successfully. So something has likely changed since then, and it likely isn't anything we've written as there isn't quite much git history since removing Travis CI: https://github.com/webpack-contrib/webpack-bundle-analyzer/commits/master

Any help in fixing our CI would be highly appreciated! :pray:

valscion (approved):
Thanks, this looks good to me!

valscion:
Released in v4.4.1, thank you for your contributions ☺️

PR #435, opened 2021-04-13
Fix tests timeout

deanshub:


valscion:
Feel free to increase the timeout to e.g. 30 seconds — if that fixes the tests sometimes timing out then we're happy.

deanshub:
referencing FYI #433 and #432

test/analyzer.js
```diff
@@ -26,7 +26,7 @@ describe('Analyzer', function () {
 
   it('should support stats files with all the information in `children` array', async function () {
     generateReportFrom('with-children-array.json');
-    await expectValidReport();
+    return expectValidReport();
```
valscion, line 29: Was this change required, or was increasing the timeout value enough? I see that there are some other specs which have the `await expectValidReport()` and they don't seem to be failing now.

test/plugin.js
```diff
@@ -15,7 +15,7 @@ describe('Plugin', function () {
 describe('Plugin', function () {
   let nightmare;
 
-  this.timeout(3000);
+  this.timeout(5000);
```
valscion, line 18: Can we play it safe and increase this to a large enough value that should not timeout in the future, either? Maybe something like 15 seconds should be safe enough?

```suggestion
  this.timeout(15000);
```

test/analyzer.js
deanshub: timeout here didn't change, just know that in some cases mocha doesn't handle await properly and the easy way is to just return the async function

deanshub: seems like it didn't solve the issue, increasing timeout

valscion: Should we then change the other `await expectValidReport` usages in this file, too? The test files should go through babel, if I'm not mistaken, and get transpiled so that `async`/`await` would work in the mocha tests...

valscion (approved):
Thanks! Let's hope this keeps the CI green ☺️

Issue #375, opened 2020-08-11
Feature Request: Allow JSON and HTML output at the same time.


leipert:
Rather than adding the plugin twice with e.g.:

```js
[
      new BundleAnalyzerPlugin({
        analyzerMode: 'static',
        openAnalyzer: false,
        reportFilename: 'webpack-report/index.html'),
      }),
      new BundleAnalyzerPlugin({
        analyzerMode: 'json',
        openAnalyzer: false,
        reportFilename: 'webpack-report/index.json'),
      }),
]
```

it would be nice to be able to generate JSON and HTML in one go, e.g.:

```js
[
  new BundleAnalyzerPlugin({
    analyzerMode: "static",
    openAnalyzer: false,
    reportOutput: [
      { type: "json", fileName: "webpack-report/index.json" },
      { type: "html", fileName: "webpack-report/index.html" },
    ],
  }),
];
```

leipert:
By the way, great work on this plugin. I don't have time right now to contribute this myself, but it would be a valuable addition to us and I just wanted to write the idea down somewhere. 😄

Gongreg:
Hey, I think it makes sense, but we should get @valscion opinion on it :)

valscion:
Thanks for the idea! I'm not sure if this works out nicely with all the possible options configurations — it looks like we could end up with configuration that is "invalid" and we would have to write quite a lot of error handling code to deal with this situation.

pkuczynski:
I believe this is possible already. Here is my example setup:

```
    plugins: [
        new BundleAnalyzerPlugin({
            analyzerMode: 'static',
            reportFilename: '../report/stats.html',
            statsFilename: '../report/stats.json',
            defaultSizes: 'gzip',
            generateStatsFile: true,
            openAnalyzer: false
        })
  ]]
```

alex-knyazev:
@

> I believe this is possible already. Here is my example setup:
> 
> ```
>     plugins: [
>         new BundleAnalyzerPlugin({
>             analyzerMode: 'static',
>             reportFilename: '../report/stats.html',
>             statsFilename: '../report/stats.json',
>             defaultSizes: 'gzip',
>             generateStatsFile: true,
>             openAnalyzer: false
>         })
>   ]]
> ```
> 
> This produces at the same time `stats.html` and `stats.json`. Is this what you wanted to achieve?

[...]

PR #442, opened 2021-05-05
Add support for writing multiple reports at once

leipert:
We are adding support for a new option called `reports` which accepts an
array of reports:

```json
new BundleAnalyzerPlugin({
  analyzerMode: "static",
  reports: [
    { type: "json", filename: "my-report.json" },
    { type: "html", filename: "my-report.html" },
  ],
});
```

Users of the `analyzerMode` in `json` should migrate from:

```js
new BundleAnalyzerPlugin({
  analyzerMode: "json",
  reportFilename: "my-report.json",
});
```

to

```js
new BundleAnalyzerPlugin({
  analyzerMode: "static",
  reports: [{ type: "json", filename: "my-report.json" }],
});
```

Users of the `analyzerMode` with `html` should migrate from:

```js
new BundleAnalyzerPlugin({
  analyzerMode: "static",
  reportFileName: "my-report.html",
});
```

to

```js
new BundleAnalyzerPlugin({
  analyzerMode: "static",
  reports: [{ type: "html", filename: "my-report.html" }],
});
```

In order to help with the migration, these config changes will pe
printed with a logger warning, while we internally being rewritten to
the new format so that this isn't a breaking change.

### DIFF
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -7,6 +7,20 @@ const viewer = require('./viewer');
 const utils = require('./utils');
 const {writeStats} = require('./statsUtils');
 
+const REPORT_TYPES = {
+  json: 'json',
+  html: 'html'
+};
+
+function beforeAfter(configBefore, configAfter) {
+  return [
+    'Change the following config parameters from:',
+    JSON.stringify(configBefore, null, 2),
+    'To these new config parameters:',
+    JSON.stringify(configAfter, null, 2)
+  ].join('\n');
+}
+
 class BundleAnalyzerPlugin {
   constructor(opts = {}) {
     this.opts = {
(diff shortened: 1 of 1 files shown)


leipert:
@valscion This is code only for now, if you are satisfied with the code changes, I will update the docs.

valscion:
I don't know if this comment by @th0r applies here: https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/375#issuecomment-828378309

I'll let @th0r communicate here if this direction is good or not.

> > Hmm yeah, I think adding support for `analyzerMode` possibly being an array would be reasonable.
> 
> Well, I think we can do that, but it will require a change to `reportFilename` option to be able to return different filenames for different report types e.g. allow to make it a function `(reportType) => <reportFilename>` (which won't work for CLI) or an object `{<reportType>: <reportFilename>}`. Anyway, options becomes a mess... We definitely need to ship a new major version with completely revised API with pluggable external reporters.

PR #384, opened 2020-10-12
Compiler.outputFileSystem.constructor of Webpack can be undefined in webpack 5.0

iamssen:
image

After webpack 5.0 update `BundleAnalyzerPlugin.getBundleDirFromCompiler()` throws errors.

`Compiler.outputFileSystem.constructor` is undefined.

After this fix, my configuration works fine in the webpack 5.0.

PR #447, opened 2021-05-13
check this.compiler.outputFileSystem.constructor is not undefined

kedarv:
via https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/384#issuecomment-838827301

In some cases, `this.compiler.outputFileSystem.constructor` can be undefined, causing access to properties on this to fail. This PR adds a check to prevent this.

### DIFF
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -134,6 +134,9 @@ class BundleAnalyzerPlugin {
   }
 
   getBundleDirFromCompiler() {
+    if (typeof this.compiler.outputFileSystem.constructor === 'undefined') {
+      return this.compiler.outputPath;
+    }    
     switch (this.compiler.outputFileSystem.constructor.name) {
       case 'MemoryFileSystem':
         return null;


valscion:
Yeah we can add tests later to get this bug fix out.

I'll copy over what I wrote in https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/384#issuecomment-842324202 to here:

> as long as there are no tests, it's possible that we'll break the fix even in a patch release as there is no way we can make sure that this fix doesn't suddenly stop working unless we have test coverage for that.

alexander-akait:
@valscion It looks like I'll have to waste time on a pretty obvious bug :confused: :disappointed:

valscion:
I'll get this bug fix out, no worries about that. I'm merely noting the effect a missing test coverage has for this kind of bug fix.

It's not possible for us to keep this bug fix working _for sure_ if we do changes to the code later. Reproducing the bug takes effort we're not willing to do every time we change the code.

valscion:
I do appreciate @alexander-akait you reproducing the original issue in #384! It was very helpful ☺️ it makes it possible for us to ship this fix even without test coverage, and I do appreciate that 💞

alexander-akait:
Can you try using this https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/384#issuecomment-838569309? Do you see the error? Just want check it too

valscion:
Yes I see the error with the reproduction

vesa_Vesas-MBP____code_muut_webpack-bundle-analyzer_issue-384-repro

valscion:
This fix has been released in v4.4.2 🎉

Issue #269, opened 2019-04-16
Add "Import Chain" into hoverstate for module.


TheLarkInn:
#### Feature description
Currently there is no way to see "why did this module end up in my bundle". It is one of the next actionable items for a developer to solve when they are triaging their JavaScript bundles. 

Today, the webpack stats object does provide this information through an API attached to the module called `issuerPath`. `module.issuerPath` will provide an array of modules that caused the current `module` to be imported or added to the bundle. 

For example if module `c` was imported by `b` and `b` was imported by `a`: The `issuerPath` for module `c` will be `[{id: a}, {id: b}]` (always starting with the entry). 

I believe that we should display this information when a developer hover's over the module in question they are curious about. 

I spent some time last Friday to explore the `./client` app which would drive the UI in the analyzer and it looks like the data is "massaged" in a way that extra data noise is removed from the "moduletree data". 

Work involved would be: add issuerPath as a property on Module on the `anaylseStats` fn/side of the app, and then include markup in the ModuleTree to incorporate this feature. 
...

valscion:
@bregenspan this might be interesting to you, as it seems to overlap a bit with https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/265

valscion:
This would be a neat feature. I wonder what would be a nice UI for that? True, showing this information in the tooltip could be a start, but we could probably do more

bregenspan:
Now that there's a context menu concept in place, maybe that could be a good place to trigger this sort of thing. (The current context menu only relates to the bundle, not the specific module selected in the bundle, but this could be changed). Something like opening context menu via right-click/Ctrl-click on a module, clicking "show reasons for inclusion", and this opening a UI panel with the `reasons`. That's a click or two more than the approach in #85 though.

th0r:
@bregenspan yes, that was one of the use cases for the context menu I was talking about in your PR ;)

TheLarkInn:
@th0r @bregenspan: 

So I also discussed this with @sokra and it turns out that the data point that should be used is `reasons` (which lives on the module object). It is not as convienent to work with (vs. `module.issuePath`) but he described to me an edge-case that would case `module.issuePath` to be inaccurate.  

> sokra [11:40]
[...]

PR #475, opened 2021-10-07
Add module reasons array

netanel8z:
Following this issue https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/269 and this branch https://github.com/webpack-contrib/webpack-bundle-analyzer/tree/module-import-reasons.
I add the reasons array to module so it will be in the analyzed report.
UI is still needs to be handle.

### DIFF
--- a/src/tree/Module.js
+++ b/src/tree/Module.js
@@ -51,7 +51,7 @@ export default class Module extends Node {
 
   toChartData() {
     const reasonsNames = _(this.data.reasons)
-      .map('module')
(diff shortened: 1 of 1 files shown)


deanshub:
Hey @valscion @th0r, 👋 
We decided to split the #269 task in 2, 
this PR adds the necessary data and if this'll get merged we'll try to make another PR to add a UI for that.

Can you review it and tell us what you think?

PR #477, opened 2021-10-11
plugin use infrastructureLogger and output the report.html to compiler.outputFileSystem

### LINKED ISSUE #476: [Nice to Have] plugin should use webpack InfrastructureLogger and output report to compiler.outputFileSystem
#### Issue description
There are 2 issues for the plugin code(not bin/analyzer). Currently, it is not a bug but nice to have
1. `BundleAnalyzerPlugin.js` should use InfrastructureLogger from webpack compiler. When I run my jest test, there are some noise message from the plugin. See the following screenshot. 
image
I can silent the `console.log` for my jest test. But I have to write some to find the plugin in my shared `webpack.config.js` and then replace the existing plugin with another one to silent log messages. 
2. When I launch the webpack-dev-server, I expect my `dist` should be empty. Currently it always has report.html. Other assets are in memfs. I have a jest test to check the folder is empty. Now I have to write more code to handle it differently
image

wood1986:
fix #476

### DIFF
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -85,7 +85,7 @@ class BundleAnalyzerPlugin {
 
   async generateStatsFile(stats) {
     const statsFilepath = path.resolve(this.compiler.outputPath, this.opts.statsFilename);
-    await this.fs.promises.mkdir(path.dirname(statsFilepath), {recursive: true});
+    utils.mkdirpSync(this.fs, statsFilepath);
 
     try {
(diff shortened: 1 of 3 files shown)


valscion:
Oh right, you removed the `jest.setTimeout` part.

th0r:
@wood1986 may I kindly ask you to:
1. Fix PR review issues in a separate commits and don't use force-push as it's really difficult to track changes you've made
2. Don't resolve my comments - I would like to do this myself.

alexander-akait:
@th0r @valscion What we need to finish it?

PR #478, opened 2021-10-11
fix: stop publishing src folder to npm

### LINKED ISSUE #472: big package size
#### Issue description

Run `npm i webpack-bundle-analyzer` and go to `node_modules/webpack-bundle-analyzer` and you will find `src` and `lib` directories, but files used only from `lib` directory, we should use `files` in `package.json` to reduce package size

#### Technical info

It does not matter

#### Debug info

It does not matter

wood1986:
fix #472

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,7 @@
   },
   "files": [
     "public",
-    "lib",
-    "src"
+    "lib"
   ],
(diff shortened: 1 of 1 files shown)


valscion:
Released in v4.5.0 ☺️

PR #501, opened 2022-03-27
stable color rendering

### LINKED ISSUE #500: FoamTree color stability
#### Issue description

It's currently impossible to compare two outputs side-by-side since the boxes can shift around and the colors are seemingly random. I'd like a way to be able to compare differences between two bundle analysis and the easiest way I can think of achieving that is to make the colors deterministic rather than random.

#### Technical info

Not really an issue, more of a suggestion.

From the FoamTree docs, we could use [groupColorDecorator](https://get.carrotsearch.com/foamtree/demo/api/index.html#groupColorDecorator) and attach the color to each item in the list of groups. (See the "Colors defined in the data model" section of the docs). To ensure stability between different runs, the colors should be based on the file name of the chunk. Webpack's chunks are in the form of `<id>.<hash>.<ext>` so we should only consider the first part of that when determining the color since the hashes will change.

I believe the id numbers of each chunk also are in a finite range (cannot confirm but just from testing) so we could translate the 1-999 ids with a percent in the `hsl` color space. This should give us an easy way to map between chunks and colors and would give a pretty decent spread of colors.

For chunks with names rather than numbers as their ids, we could come up with a similar hashing algorithm which turns the characters into a color.

#### Debug info

N/A

CreativeTechGuy:
Fixes #500 

There's a ton of variance in how users may configure their chunk names. To ensure stable colors, we need to dynamically identify which (if any) part of the file name is the unique "name" part and use that to determine the colors. If that cannot be identified or if a file doesn't follow the pattern of the rest, the entire file name will be used.

Given that the behavior before was random rainbow colors all the time (colors would even change when zooming/hiding) this is better even in the case where nothing can be uniquely identified. So I'd argue that while this is an imperfect solution, it's better in all cases than what was there by default.

-----

This is best visualized with photos of the different cases:

Before this PR:

old

After this PR (when unique name can be identified):

new-before-change

After this PR (when unique name can be identified - with chunk contents changed but colors are still stable with above):

new-with-change

After this PR (with no identifiable parts of the filename - deterministic but effectively random colors):

no-names


valscion:
Thanks for opening a PR! This is indeed an interesting approach.

The loops look a tad scary to me, as I don't know how they would perform with large graphs. I also am not sure if there is no way an infinite loop could happen there.

Is there a way to get some data to the graph data object itself that could be used for this purpose with less likelihood for a performance hit?

CreativeTechGuy:
Thanks for taking the time to review @valscion!

Quick time complexity analysis just to make sure we are on the same page. I assume you are talking about `findChunkNamePartIndex`.

It is almost entirely bounded by the number of chunks, secondarily the number of parts of the most complex file name.

So if there are 100 chunks and the most complex file name is `a.b.c.d.e.f.g.h.i.j.k.l.m.n.o.p.q.r.s.t.u.v.w.x.y.z.js` (27 parts) then it'd be:

```
O(100 * 27) = O(2700)
```

The only thing I really see as performance impacting is if a user has a ton of chunks, realistically that's the only variable that affects time complexity in a realistic way here. But if someone has a few hundred chunks, the canvas rendering will also be impacted by having to compute the layout of all of those chunks too. I'd suspect this function's additional performance hit wouldn't be noticeable in comparison.

------

The only place I can see an infinite loop being possible is in `getGroupRoot`, but that tree traversal is very similar to the one just a few lines below in `zoomToGroup` which seems to be working fine. Assuming that an infinite loop isn't possible here, I believe the number of nested groups in the graph is based on the folder structure, so I'd be surprised if there's any node that is more than a dozen folders deep in the project.

------

Let me know if we are on the same page here and if you want me to make any edits or run any tests to make sure this is working correctly at scale.

valscion:
Thanks for the additional information 😊 good point about the count of chunks impacting the layout calculation anyway.

I have been able to only glance at the code on mobile so far so I haven't yet been able to do a more thorough review.

CreativeTechGuy:
@valscion have you been able to take a look? This seems like a valuable improvement that would be great to get merged! :)

valscion:
Sorry, this is still pending. I'm currently on paternal leave so I have little time to spend on computer 😊

valscion:
Hi @CreativeTechGuy! I'm still on parental leave, so I haven't yer had a chance on checking this out. I haven't forgotten about this PR, though, and will get back to this once I have more time ☺️. Thank you for your patience

valscion:
The only consideration I have in here is that the choice of colors now is quite bright and hides some details in some places that the FoamTree built-in color algorithm was able to discover:

Screenshot_2022-08-17_at_10-24-06_venuu_17_Aug_2022_at_10_23

Compare that to the original version (the chunk contents are the same) and you can see that there are more details immediately visible before this PR:

Screenshot_2022-08-17_at_10-24-00_venuu_17_Aug_2022_at_10_21

So I wonder is there a way to tweak the color deciding algorithm to get a bit more muted colors so that they'd be easier on the eye and also be able to see the count of sub-modules more easily?

CreativeTechGuy:
I'd argue that the colors in the original version are just as similar (see the huge cluster of similar teals in the bottom right) but it is just more noticeable now that the colors are more vibrant. I'm not sure there's simply enough colors which are visually distinct enough to avoid this problem.

I'll work on making the colors more muted. Although I think that the more muted the colors are, the more that will start to look visually similar.

In my testing, as we reduce the saturation on the colors, it starts to look "disabled" so I don't want to mute them too much that it starts to communicate some functional state which would be misleading.

CreativeTechGuy:
I toned down the saturation a bit. I think this is a good middle ground. It looks like this 
image

Hopefully you like it!

valscion:
> I toned down the saturation a bit. I think this is a good middle ground. It looks like this

Oh wow this looks much nicer!

I see no reason why we would not be able to ship this soon. Do you think this feature warrants a small mention in the README.md or is it OK to leave undocumented?

Adding a changelog entry at least would be nice ☺️. Make sure to update your branch (merge commit is OK to me) with the `master` branch in this repository before tweaking the changelog as otherwise there's likely gonna be conflicts.

CreativeTechGuy:
Got it! Rebased with master and added a line to the Changelog!

I'm not sure about documenting this. It's not configurable and shouldn't break anything. Also, if someone doesn't understand it and just thinks they are random, then there's no harm. Worst case it's the same as it was previously, best case it's an improvement. :)

valscion (approved):
Thanks! This looks great to me! :+1:

valscion:
This has now been released in v4.6.0 ☺️

PR #505, opened 2022-03-31
Reduce the usage of lodash

SukkaW:
Why do we need lodash if we can utilize features provided by the ECMAScript itself :)


tgmarinho (approved):

src/tree/utils.js
```diff
@@ -1,20 +1,14 @@
...
-
 const MULTI_MODULE_REGEXP = /^multi /u;
 
 export function getModulePathParts(moduleData) {
   if (MULTI_MODULE_REGEXP.test(moduleData.identifier)) {
     return [moduleData.identifier];
   }
 
-  const parsedPath = _
-  // Removing loaders from module path: they're joined by `!` and the last part is a raw module path
```
valscion, line 11: These comments are important as they document the reason why the lines of code exist in the first place. Could we keep them?

valscion (commented):
Yeah seems like these changes end up having the same outcome as the lodash code. Can you address my review comment and also do a changelog entry with an "Internal" tag?

src/tree/utils.js
SukkaW: Sure, will do!

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 * **Improvement**
   * Display modules included in concatenated entry modules on Webpack 5 when "Show content of concatenated modules" is checked ([#602](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/602) by [@pgoldberg](https://github.com/pgoldberg))
 
+* **Internal**
+  * Replace some lodash usages with JavaScript native API ([#505](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/505)) by [@sukkaw](https://github.com/sukkaw).
+
```
valscion, line 22: The new changelog entry should be between `## UNRELEASED` and `## 4.9.0` ☺️

valscion (changes requested):

valscion:
Looks like some tests are breaking. Have you tried if running tests locally pass for you?

SukkaW:
> Looks like some tests are breaking. Have you tried if running tests locally pass for you?

Yeah, I have been looking into this. However, it only shows `timeout` (instead of test mismatch) on my side.

valscion:
Looks like there are still some test failures only found in this PR and also lint errors.

SukkaW:
> Looks like there are still some test failures only found in this PR and also lint errors.

Fixed the `viewer.js` test in [`f162c20` (#505)](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/505/commits/f162c20b5cc5beee362ae01f9cdf6b8618114411) and fix lint errors from `npm run lint` in [`ead9a9b` (#505)](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/505/commits/ead9a9b729905e97975808d74283b4545f477c03).

valscion (approved):
Thanks, looks good to me! The diff is small enough that it seems like it shouldn't cause any regressions.

This doesn't look like it needs to be urgently released so I'll give it some time until there are other things that warrant a new release ☺️

SukkaW:
> Thanks, looks good to me! The diff is small enough that it seems like it shouldn't cause any regressions.
> 
> This doesn't look like it needs to be urgently released so I'll give it some time until there are other things that warrant a new release ☺️

And I am planning more PRs (that I want to be included in the next release) as well.

PR #520, opened 2022-08-16
feat: support custom loginfo with server mode

southorange1228:
Sometimes,i will set host to be `0.0.0.0`,but i won't console info with `0.0.0.0`.So i need to custom log info and show better info for users.


valscion:
Hi, thanks for the pull request! How are you intending to use this change in your application?

southorange1228:
> 

Actually, i create a cli tool for my team.

example for use

```js
const BundleAnalyzerPlugin = require("webpack-bundle-analyzer").BundleAnalyzerPlugin;
// webpack.config.js
plugin: [
  new  BundleAnalyzerPlugin({
    host: '0.0.0.0',
    logInfo: () => {
      // some other logic
      return 'custom log info';
    }
  })
]
```

I will add analytics in cli tool. If i can use it like this demo.I think it's fine

src/utils.js
```diff
@@ -53,6 +53,15 @@ exports.defaultTitle = function () {
   return `${process.env.npm_package_name || 'Webpack Bundle Analyzer'} [${currentTime}]`;
 };
 
+exports.defaultLogInfo = function (options) {
+  const {host, port} = options;
+  const url = `http://${host}:${port}`;
+  return (
+    `${bold("Webpack Bundle Analyzer")} is started at ${bold(url)}\n` +
+    `Use ${bold("Ctrl+C")} to close it`
+  );
```
valscion, line 62: This now duplicates the "Webpack Bundle Analyzer is started at..." text by default, as it's also written here:

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/45b4db76dff3f3b21cea5ae7f9859c60f2cb0f7d/src/viewer.js#L82-L85

I wonder if your use case could also be implemented by only changing how the `url` is calculated? So that we would not need to allow users of this plugin to also override the "Webpack Bundle Analyzer is started" and "Use Ctrl+C to close it" texts.

Then we could change the name of this option to not be `logInfo` but something like `analyzerUrl`, and usage would look like:

```js
const url = analyzerUrl({
  listenPort: port,
  listenHost: host,
  boundAddress: server.address()
});
```

We should document that the `server.address()` is the value as returned by `http` module: https://nodejs.org/api/net.html#serveraddress

The default implementation should be what is the current returned value:

```js
exports.defaultAnalyzerUrl = function (options) {
  const { listenHost, boundAddress } = options;
  return `http://${listenHost}:${boundAddress.port}`;
}
```

valscion (changes requested):

src/utils.js
southorange1228: copy that! I fixed. Any other suggestions？

valscion:
> copy that! I fixed. Any other suggestions？

If you could add a changelog entry to `CHANGELOG.md` then I think this is OK. You will also need to sign the CLA like the bot commented above:

https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/520#issuecomment-1216212707

southorange1228:
> > copy that! I fixed. Any other suggestions？
> 
> If you could add a changelog entry to `CHANGELOG.md` then I think this is OK. You will also need to sign the CLA like the bot commented above:
> 
> [#520 (comment)](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/520#issuecomment-1216212707)

OK,i finished!

valscion:
OK so the usage could look something like this, right?

```js
plugins: [
    new  BundleAnalyzerPlugin({
      analyzerUrl: ({ listenPort, listenHost, boundAddress }) => {
        return `http://some-domain.example.com:${boundAddress.port}`;
      }
  })
]
```

Can we add a short documentation to the README.md about this option, too? https://github.com/webpack-contrib/webpack-bundle-analyzer#options-for-plugin

It could be below the `analyzerPort` value there.

> Name: **`analyzerUrl`**
> Type: `{function}`, called with `{ listenHost: string, listenHost: string, boundAddress: server.address}`.  [`server.address` comes from Node.js](https://nodejs.org/api/net.html#serveraddress)
> Default: `http://${listenHost}:${boundAddress.port}`. The URL printed to console with `server` mode

southorange1228:
Alright, it's my first time to contribute to this project and thanks for your guide.Thanks a lot ：）

valscion:
Thank you for your contributions ☺️ this PR seems good to go to me, hopefully it be useful to others, too ☺️

valscion (approved):

valscion:
This has now been released in v4.6.0 ☺️

PR #521, opened 2022-08-19
Update some devDependencies

valscion:
A lot of updates have happened since the last time packages have been updated. I ignored some packages which seemed like they'd be more effort to upgrade to right now to make this PR easier to handle.

This PR also drops support for Node.js v10 as some devDependencies no longer work on it.


valscion:
Huh, why didn't the tests start to run.... baaaah.

PR #526, opened 2022-09-01
Remove david-dm from README

renbaoshuo:
The `david-dm.org` website is down for 3 months, and the project `alanshaw/david` is not actively maintained anymore. `david-dm.org` should be considered dead, thus badges should be removed from README.


snitin315 (approved):

valscion (approved):
Thanks!

PR #549, opened 2023-01-10
feat: add strictPort options

PengBoUESTC:
add `strictPort` option, if current port is in use, set strictPort to `false` so that this plugin try to restart with next port!


valscion:
Hi, thanks for opening a PR!

Is there a reason why `analyzerPort: auto` won't work for you? We already have that option so I'd rather avoid having two ways to achieve a similar result.

PengBoUESTC:
> analyzerPort

Actually , i didn't realize the meaning of `analyzerPort: auto` ,  maybe there should be some explanation of this value(`auto`) like this: 

If analyzerPort is `auto`, the operating system will assign an arbitrary unused port, 

> If port is omitted or is 0, the operating system will assign an arbitrary unused port, which can be retrieved by using server.address().port after the ['listening'](https://nodejs.org/dist/latest-v18.x/docs/api/net.html#event-listening) event has been emitted.

thanks a lot

PR #412, opened 2021-01-19
Upgrade to FoamTree 3.5.0, now available as an NPM package.

stanislawosinski:
The latest FoamTree releases fix various functional bugs and performance regressions in the flattened display mode used by webpack-bundle-analyzer.

PR #554, opened 2023-01-30
feat: gray out non-matches when searching

### LINKED ISSUE #553: Search should gray out non-matches?
#### Issue description

When searching for bundles, the current scheme seems to be to show the matches in red transparent overlay, but leave all of the rest of the colors the same. This makes it fairly hard to see what bundles matched: 1) lots of colors, some quite close to red (e.g. dark magenta); 2) since the match overlay has a transparency, this causes the matches to be even closer to the almost-reds; 3) if there are lots of small matches (which add up), this doesn't jump out.

Suggestion: gray out the non-matches?

without search:
without search

with search string "monaco":
with search

#### Technical info

<pre>
  System:
    OS: macOS 13.2
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 14.75 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.19.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.3 - /usr/local/bin/npm
</pre>

#### Debug info

As a webpack plugin.

No options.

It would be nice to also attach webpack stats file.
It can be generated using these options:
```js
new BundleAnalyzerPlugin({
  analyzerMode: 'disabled',
  generateStatsFile: true,
  // Excludes module sources from stats file so there won't be any sensitive data
  statsOptions: { source: false }
})
```
`stats.json` will be created in Webpack bundle output directory.

starpit:
This PR reduces the saturation (from 60 to 10) for non-matches.

Note: there seems to be a bug in foamtree? It does not redraw some of the nodes until you hover over them. This seems to be a preexisting issue.

Fixes #553

Screenshot 2023-01-30 at 11 00 33 AM

<details><summary>Click to see the outdated details about the <code>.redraw()</code> behavior we've discussed in this PR</summary>

## FoamTree bug?

This gif shows how randomly some of the nodes are not redrawn (by foamtree) until you hover over them. I think this bug is masked by the current highlighting behavior, which does not change the color of non-matching nodes.
foamtree-oddity

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
  * **Improvement**
    * Support reading large (>500MB) stats.json files ([#423](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/423) by [@henry-alakazhang](https://github.com/henry-alakazhang))
+   * Improve search UX by graying out non-matches ([#554](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/554) by [@starpit](https://github.com/starpit))
 
 ## 4.7.0
 
--- a/client/components/Treemap.jsx
+++ b/client/components/Treemap.jsx
@@ -22,11 +22,7 @@ export default class Treemap extends Component {
         dataObject: this.getTreemapDataObject(nextProps.data)
       });
     } else if (nextProps.highlightGroups !== this.props.highlightGroups) {
-      const groupsToRedraw = [
-        ...nextProps.highlightGroups,
-        ...this.props.highlightGroups
-      ];
-      setTimeout(() => this.treemap.redraw(false, groupsToRedraw));
+      setTimeout(() => this.treemap.redraw());
     }
   }
 
@@ -102,6 +98,11 @@ export default class Treemap extends Component {
             b: 0,
             a: 0.8
           };
+        } else if (highlightGroups && highlightGroups.size > 0) {
+          // this means a search (e.g.) is active, but this module
+          // does not match; gray it out
+          // https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/553
+          variables.groupColor.s = 10;
         }
       },
       /**


valscion:
> This gif shows how randomly some of the nodes are not redrawn (by foamtree) until you hover over them. I think this bug is masked by the current highlighting behavior, which does not change the color of non-matching nodes.

Hmm, if this bug wasn't active earlier but is now visible, we will need to fix it before we can ship this feature.

Are you able to debug further on why this happens and if updating our version of FoamTree would help fix the situation?

Ever since https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/412 we've got FoamTree from an NPM package. I don't recall right now on how updating this package will happen so it needs some thought behind it. Looking through the changelogs of FoamTree could show if this bug has existed on FoamTree side and has now been fixed.

starpit:
Hmm, I don't think FoamTree is open source? I don't even see a way to file an issue.

valscion:
Yeah FoamTree is not open source but they do distribute it via npm

starpit:
Interesting. If i change the `.redraw(false, groupsToRedraw)` to `.redraw()`, then the bug goes away. So either: a) the `groupsToRedraw` computation is incorrect on your side; or b) foamtree does not handle this properly.

Is it viable to work around this for now by always redrawing all groups, i.e. `.redraw()`?

valscion:
Thanks for looking more into why this case happened 😊. I'm not that familiar with the treemap interaction code so it is very helpful to have you figure this out, too.

It sounds like the original code could've used the `false` argument as some sort of render optimization. Is the rendering now noticeably slower than it used to be? I wonder if we need to somehow have some sort of a debounce for the handling of search input value changes if redrawing the treemap becomes slow.

starpit:
the `false` of `.redraw(false)` is the default behavior. it tells foamtree *not* to animate the update. the second argument is a hint to foamtree as to which groups may have changed. the bug is either in bundle analyzer's presentation of changed groups to foamtree, or in foamtree's algorithm that interprets this hint. 

i have not noticed any rendering lag for the example above, which has maybe 100 top-level groups, and 600 total groups.

valscion:
Okay thanks for the extra information!

That sounds like a useful optimization to have. I suspect that operation is costly so with users having even more groups and maybe a less powerful machine could get performance issues they weren't seeing earlier.

However, we are indeed debouncing changes done to the search input here:

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/38930418af195b5ecbc857ce1db43bf83256d726/client/components/Search.jsx#L42-L44

so maybe the performance issue wouldn't be that bad. I don't know what other code paths call to this method that now calls `.redraw()`? I hope none related to cursor movements or some other commonly used events.

valscion (approved):
Ok I took this for a spin, too, and even with a CPU slowdown of 6x I can't see any performance issues with this code.

Thank you for the contribution, this looks ace! I'll create a changelog entry for this and merge after ☺️

valscion:
This is now in v4.8.0! 🚀

PR #602, opened 2023-05-30
improvement: Display more detailed module information for concatenated entry modules

pgoldberg:
After migrating a large monorepo from Webpack 4 to Webpack 5, we noticed that we lost detailed entry module information, which made the bundle analyzer reports much less useful for us. After looking a little bit through the `webpack-bundle-analyzer` repo, I found this comment: https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/e231c77064329fdd586e2e69f65d24f9a5a2004a/src/analyzer.js#L124-L126

Although we can't determine the parsed sizes of the concatenated modules, that shouldn't prevent us from displaying what modules are actually included in the concatenated entry module when viewing stat sizes. This is what our current report looks like for entry modules (with Stat sizes selected and "Show content of concatenated modules" checked):

image

After this change, here is what that same view looks like:

image

### DIFF
--- a/src/tree/ConcatenatedModule.js
+++ b/src/tree/ConcatenatedModule.js
@@ -38,7 +38,8 @@ export default class ConcatenatedModule extends Module {
       currentFolder = childFolder;
     });
 
-    const module = new ContentModule(fileName, moduleData, this);
+    const ModuleConstructor = moduleData.modules ? ConcatenatedModule : ContentModule;
+    const module = new ModuleConstructor(fileName, moduleData, this);
     currentFolder.addChildModule(module);
   }
 


valscion:
Huh, that's a neat change with such a small change! Are you able to write up some sort of test to verify this behavior?

pgoldberg:
@valscion I added a test, and also updated the PR to include the modules contained in the concatenated entry module's graph for the parsed and gzipped reports, too.

After thinking a bit more about this I think this is actually a bug and a regression in `webpack-bundle-analyzer` from Webpack 4 to Webpack 5. On Webpack 4, entry modules that were themselves concatenated modules still showed (inaccurate) parsed and gzipped sizes when that box was ticked.

On Webpack 5, because we can't determine the exact parsed size of the entry modules, we create a concatenated module containing the entry modules. For those modules, we do get estimated parsed and gzipped sizes, because they're all `ContentModules`.

However, the entry modules themselves can be concatenated modules. In that case, because we've created the module as a `ContentModule` rather than a `ConcatenatedModule`, we're not getting estimated sizes for the modules within the concatenated entry module.

I believe this bug fix should only affect entry modules, because this type of situation only really occurs when an entry module is a concatenated module. I don't think webpack would ever output a real concatenated module that is itself a concatenated module – this only happens because of the synthetic concatenated module we create of all the entry modules that we can't accurately parse.

The test I added was generated using this minimal repro I made: https://github.com/pgoldberg/webpack-bundle-analyzer-concatenated-entry-modules

src/tree/ConcatenatedModule.js
```diff
@@ -14,6 +14,22 @@ export default class ConcatenatedModule extends Module {
...
+    return this.getGzipSize() ?? this.getEstimatedSize('gzipSize');
+  }
+
+  getEstimatedSize(sizeType) {
+    const parentModuleSize = this.parent[sizeType];
+
+    if (parentModuleSize !== undefined) {
+      return Math.floor((this.size / this.parent.size) * parentModuleSize);
+    }
+  }
```
pgoldberg, line 31: This is taken from here: https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/a4c304e6c2d1c59a57dde6e747f9ce8e4b46a077/src/tree/ContentModule.js#L18-L24

valscion (commented):
Looks solid to me!

I have one wish regarding the example case that if possible, would make debugging this behavior later somewhat easier.

Also, could you update the changelog to include this change?

valscion (approved):
Thanks, looks great!

pgoldberg:
@valscion Thanks for the quick review! 😄

If you could cut a release whenever you get a chance, I'd really appreciate it 🙂

valscion:
Yeah ran out of time yesterday, I usually cut releases almost straight away. I'll do a new release now :)

valscion:
Released now in v4.9.0! ☺️

pgoldberg:
Thank you!!

PR #609, opened 2023-08-02
Reduce installation size

SukkaW:
The PR follows #505. The PR reduces `node_modules` size after `npm i webpack-bundle-analyzer` with 2 changes:

- Replaces `chalk` with `picocolors`
- Replaces `lodash` with individual `lodash.*` packages.

**Comparison**

Running `npm i --omit=dev --omit=optional` against `webpack-contrib/webpack-bundle-analyzer:master`:

image

Running `npm i --omit=dev --omit=optional` against `sukkaw/webpack-bundle-analyzer:optimize-install-size`:

image


src/analyzer.js
```diff
@@ -179,24 +183,22 @@ function readStatsFromFile(filename) {
...
 }
 
 function getBundleModules(bundleStats) {
-  return _(bundleStats.chunks)
-    .map('modules')
-    .concat(bundleStats.modules)
-    .compact()
-    .flatten()
-    .uniqBy('id')
-    // Filtering out Webpack's runtime modules as they don't have ids and can't be parsed (introduced in Webpack 5)
```
valscion, line 197: Please keep these documentation comments as they give reasoning why certain logic exists :)

valscion (commented):
Nice :+1:. Can you add more changelog entries as well?

src/analyzer.js
SukkaW: Ah! I keep forgetting this!

valscion (approved):
Yeah looks good to me :+1:

valscion:
This PR and #505 have been released in v4.9.1 now ☺️. Thank you for your contributions!

stof:
the usage of those individual packages is discouraged: https://lodash.com/per-method-packages

Also, as soon as another package also depends on lodash, this change actually increases the size of node_modules as lodash was deduplicated before.

valscion:
Thanks @stof. So we should probably revert these changes then. @SukkaW would you be up for that?

SukkaW:
> Thanks @stof. So we should probably revert these changes then. @SukkaW would you be up for that?

For most of my projects, the only copy of `lodash` was introduced by the webpack-bundle-analyzer.

Since lodash is essentially dead (hasn't been maintained for years), IMHO instead of reverting back to lodash, replacing all `lodash.*` packages with maintained, performant, and lightweight packages would be a better approach.

stof:
Well, lodash is indeed kinda dead since 3 years. But the per-method packages have last been released 7 years ago. So they are still missing 4 years of work in the lodash package.

PR #627, opened 2024-04-07
Remove `is-plain-object`

SukkaW:
The PR removes a dependency `is-plain-object`, further reducing the installation size.

Ref #505 #609 #612

Turns out that the return value of `analyzer.getViewerData()` is always an array (from `Object.entries`), we can safely remove the `is-plain-object` guard.


valscion (approved):
Looks good to me! 👍

valscion:
This has now been released as part of v4.10.2

PR #628, opened 2024-04-08
feat: added flags

info-arnav:
The feature adds flags, which can be updated in the cli repo.

### DIFF
--- a/bin/bundle-analyzer-flags.js
+++ b/bin/bundle-analyzer-flags.js
@@ -1,0 +1,159 @@
+"use strict";
+
+module.exports = {
+  version: {
+    configs: [
+      {
+        type: "boolean",
+        multiple: false,
+        description: "Output the version number",
+        path: "version",
+      },
+    ],
+    description: "Output the version number",
+    simpleType: "boolean",
+    multiple: false,
+  },
+  mode: {
+    configs: [
+      {
+        type: "enum",
+        values: ["server", "static", "json"],
+        multiple: false,
+        description: "The mode to run the analyzer in: server, static, or json.",
+        path: "mode",
+        defaultValue: "server",
+      },
+    ],
+    description: "The mode to run the analyzer in: server, static, or json.",
+    simpleType: "string",
+    multiple: false,
+  },
+  report: {
+    configs: [
+      {
+        type: "string",
+        multiple: false,
+        description: 'Path to bundle report file that will be generated in "static" mode.',
+        path: "report",
+        defaultValue: "report.html",
+      },
+    ],
+    description: 'Path to bundle report file that will be generated in "static" mode.',
+    simpleType: "string",
+    multiple: false,
+  },
+  title: {
+    configs: [
+      {
+        type: "string",
+        multiple: false,
+        description: "String to use in title element of html report.",
+        path: "title",
+        defaultValue: "Webpack Bundle Report",
+      },
+    ],
+    description: "String to use in title element of html report.",
+    simpleType: "string",
+    multiple: false,
+  },
+  size: {
+    configs: [
+      {
+        type: "enum",
+        values: ["stat", "parsed", "gzip"],
+        multiple: false,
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
And now let's use them here - https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/src/bin/analyzer.js#L27

there is an API - https://github.com/webpack/webpack/blob/main/lib/cli.js#L612
So:
- you `require` options from this file
- then use `processArguments`
- then use `.option(...)` to create all options

info-arnav:
> And now let's use them here - https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/master/src/bin/analyzer.js#L27
> 
> there is an API - https://github.com/webpack/webpack/blob/main/lib/cli.js#L612 So:
> 
> * you `require` options from this file
> * then use `processArguments` (don't forget to handle `Problem`)
> * then use `.option(...)` to create all options

on it

src/bin/analyzer.js
```diff
@@ -25,53 +28,58 @@ const program = commander
                      By default a directory of stats file is used.`
   )
   .option(
```
alexander-akait, line 30: Will be great to reuse them here too

valscion:
What are flags?

src/bin/analyzer.js
```diff
@@ -1,21 +1,24 @@
 #! /usr/bin/env node
 
-const {resolve, dirname} = require('path');
+const { resolve, dirname } = require("path");
```
valscion, line 3: Why do we need to have so much formatting changes here?

package-lock.json
valscion, line 1: Do we need to change this big amount of packages that GitHub even has trouble visualizing the changes in `package-lock.json`?

alexander-akait:
@valscion We need these flags to support webpack CLI - https://github.com/webpack/webpack-cli/issues/1838, so you can use `webpack build --analyze --analyze-host 127.0.0.1`, so you don't need to have own CLI anymore (and can drop it in the future major release) and we will provide a better and unified API for our developers, less maintance problems for you in future too

valscion:
Thanks for the background information! It would indeed be valuable to allow this tool to be used easily through webpack CLI.

In order for me to eventually be able to review this PR, I'd need to see the least amount of changes possible. That includes:

* Not doing any unnecessary formatting changes
* Not touching dependencies in this PR (so no changes to `package-lock.json`)

And then we'd need these:

* Test coverage in here to verify we don't regress this feature in the future
  * The best test coverage would be one that would also use the webpack CLI so that we know the interplay between this package and webpack CLI works as intended.
  * I'm open to hearing alternatives if said test coverage is too difficult to add
* A PR description which outlines how this feature would is intended to work
* A changelog entry
* Maybe even a README.md change if such a change is warranted?

Also, I'd like to know whether this change will indeed still be backwards-compatible (that is, releasing this would be a minor version release and not a major release). If there is no other way than doing a major release then I'm also OK with that.

All in all, I do think that this feature would make sense — but I also have some wishes as the maintainer of this project before I can accept the changes. ☺️

alexander-akait:
@valscion Yeah, I will ping you when it will be ready to review

valscion:
Sounds nice! Thanks!

info-arnav:
@alexander-akait shall i proceed to write the tests ?

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 ## UNRELEASED
```
evenstensberg, line 14: uncommit

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webpack-bundle-analyzer",
-  "version": "4.10.2",
+  "version": "4.10.3",
```
evenstensberg, line 3: no new version

PR #640, opened 2024-10-11
upgrade package-lock.json

nektro:
ran `npm install` with v10.8.3 from node v22.9.0


valscion:
The package-lock.json is now updated after running `npm install` with a modern node version.

Issue #642, opened 2024-10-21
Dark mode optional


dsshard:
Hi! This is a super cool product, and often helps me out. Thank you!

I really miss the dark mode, since I often work at night, and the abrupt switch to the "light" tab is very irritating to the eyes.

I made a patch for myself that simply makes the body background #000 and it helps me. But it would be cool to add this as an option for the settings.

Thanks again for the product, you are awesome!

dsshard:
https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/643

valscion:
Yeah a dark mode could be useful. I'd rather avoid adding a new option for it, though — but some sort of a UI feature for a dark mode could be useful.

I could imagine that it'd work like this:

1. Check [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) if a light/dark mode setting has been previously selected. If so, use that selection.
2. Check [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) and use that as a default
3. Allow user to override what `prefers-color-scheme` had and persist that to `localStorage`

The simplest first iteration could be to only add support for `prefers-color-scheme` and not mess with `localStorage` at all.

Nitin-kumar-yadav1307:
Hi! I am a beginner and would like to work on implementing the optional dark mode feature as part of my first open-source contribution. Could you please assign this issue to me?

evenstensberg:
Can't assign you but feel free to pick this up

valscion:
Looks like it was possible to assign this issue to @Nitin-kumar-yadav1307 after all! I'll keep this one for you for a while, let us know if you hit obstacles @Nitin-kumar-yadav1307

Nitin-kumar-yadav1307:
Thanks a lot for assigning this issue to me! .I’ve successfully set up the project locally and started implementing the Dark Mode toggle in the sidebar. I’ve added the feature and rebuilt the project  however, I can’t yet see the dark mode toggle reflected in the running app. I’m currently debugging to identify what might be preventing it from showing up. Please let me know if there’s any specific setup or component I should check to ensure the toggle renders correctly.

evenstensberg:
Please don't use ai when communicating. Thanks.

valscion:
@Nitin-kumar-yadav1307 I've unassigned you now since two weeks have passed and there hasn't been much new progress to show here.

[...]

PR #643, opened 2024-10-21
feat: Add dark theme

dsshard:
Add simple dark theme
- Update readme
- Add optional param for webpack plugin
- Add cli flag


valscion:
Thanks for the PR and the idea! See https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/642#issuecomment-2476223436 for how dark/light mode option could work. Let's keep the design discussion in there and once there is code to look at, I can review it when I have a good time slot for it.

PR #646, opened 2024-11-25
Flags for Webpack-bundle-analyser

info-arnav:
This is with reference to https://github.com/webpack/webpack-cli/pull/4141

Added command line flags for the bundle analyzer to add it to Webpack-cli

### DIFF
--- a/bin/bundle-analyzer-flags.js
+++ b/bin/bundle-analyzer-flags.js
@@ -1,0 +1,165 @@
+"use strict";
+
+module.exports = {
+  version: {
+    configs: [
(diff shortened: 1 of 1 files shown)


valscion:
Thanks! Can you check if there's a way to test somehow that `webpack-bundle-analyzer --help` contains same list of items as this new file? Would be nice to avoid regressing this addition in the future.

Is there a way to test that this file is a valid options file for webpack-cli as well? I'd already be happy to have a test showing that webpack-cli doesn't crash/error with this file

Issue #556, opened 2023-02-17
Error generating bundle report: Error parsing bundle asset: 'import' and 'export' may appear only with 'sourceType: module'


eamodio:
#### Issue description
When using the analyzer to generate a report, some of my entry points work, but others report the following errors:
```
Error parsing bundle asset:  'import' and 'export' may appear only with 'sourceType: module'```

#### Technical info

<!--
Please include the output of running the command below:

npx envinfo --system --binaries --npmPackages='*webpack*' --npmGlobalPackages=webpack-bundle-analyzer 

-->

<pre>
  System:
    OS: Windows 10 10.0.22623
    CPU: (24) x64 AMD Ryzen 9 3900X 12-Core Processor
    Memory: 14.23 GB / 31.92 GB
  Binaries:
    Node: 18.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.3.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    clean-webpack-plugin: 4.0.0 => 4.0.0
    copy-webpack-plugin: 11.0.0 => 11.0.0
    csp-html-webpack-plugin: 5.1.0 => 5.1.0
    css-minimizer-webpack-plugin: 4.2.2 => 4.2.2
    fork-ts-checker-webpack-plugin: 6.5.2 => 6.5.2
    html-webpack-plugin: 5.5.0 => 5.5.0
    image-minimizer-webpack-plugin: 3.8.1 => 3.8.1
    terser-webpack-plugin: 5.3.6 => 5.3.6
    webpack: 5.75.0 => 5.75.0
    webpack-bundle-analyzer: 4.8.0 => 4.8.0
    webpack-cli: 5.0.1 => 5.0.1
    webpack-node-externals: 3.0.0 => 3.0.0
    webpack-require-from: 1.8.6 => 1.8.6
</pre>

#### Debug info
```js
new BundleAnalyzerPlugin({
	analyzerMode: 'static',
	generateStatsFile: true,
	openAnalyzer: false,
	reportFilename: path.join(out, 'webview-bundle-report.html'),
	statsFilename: path.join(out, 'stats.json'),
})
```

What other Webpack plugins were used?
[...]

PR #649, opened 2025-01-08
Fixes #427, #556 use correct `sourceType` when parsing bundles

### LINKED ISSUE #427: support webpack5 new library type 'module'
#### Issue description

Currently, It cannot parse properly when webpack5 config library type to `module`. Because webpack5 append some codes like `export {a, b, c}` to the end of the bundle.

```ts
const config: webpack.Configuration = {
  output: {
    libraryTarget: 'module',
  },
  experiments: {
    outputModule: true,
  },
  plugins: [new BundleAnalyzerPlugin({ analyzerMode: 'static' })],
}
```

#### Technical info

<pre>
  System:
    OS: macOS 11.2.3
    CPU: (4) x64 Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
    Memory: 3.01 GB / 24.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 15.11.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.6.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    case-sensitive-paths-webpack-plugin: 2.4.0 => 2.4.0 
    fork-ts-checker-webpack-plugin: 6.2.0 => 6.2.0 
    webpack: 5.28.0 => 5.28.0 
    webpack-bundle-analyzer: 4.4.0 => 4.4.0 
    webpack-cli: 4.5.0 => 4.5.0 
    webpack-merge: 5.7.3 => 5.7.3 
</pre>

#### Debug info

plugin

`new BundleAnalyzerPlugin({ analyzerMode: 'static' })`

eamodio:
Refs #427, #556

We are now parsing bundles as ES modules based on stats JSON information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 ## UNRELEASED
 
+* **Improvement**
+  * Parse bundles as ES modules based on stats JSON information
+
 ## 4.10.2
 
 * **Bug Fix**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 ## UNRELEASED
 
 * **Improvement**
-  * Parse bundles as ES modules based on stats JSON information
+  * Parse bundles as ES modules based on stats JSON information ([#649](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/649) by [@eamodio](https://github.com/eamodio))
 
 ## 4.10.2
 
--- a/src/analyzer.js
+++ b/src/analyzer.js
@@ -77,7 +77,7 @@ function getViewerData(bundleStats, bundleDir, opts) {
       let bundleInfo;
 
       try {
-        bundleInfo = parseBundle(assetFile);
+        bundleInfo = parseBundle(assetFile, {sourceType: statAsset.info.javascriptModule ? 'module' : 'script'});
       } catch (err) {
         const msg = (err.code === 'ENOENT') ? 'no such file' : err.message;
         logger.warn(`Error parsing bundle asset "${assetFile}": ${msg}`);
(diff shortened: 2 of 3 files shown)


evenstensberg (commented):
Sign the CLA please 🥰

eamodio:
Done 😀

valscion:
Thanks! I wonder if there is any other solution than forcing the user to change an option? Is there some way we could infer whethet the source content is a module and not a script?

I'd rather avoid adding more options to support something which should work out-of-the-box.

valscion:
> Each assets has `javascriptModule: true` in `info` when it is a ECMA module, so it can be infer from it (https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/index.js#L490), I don't think using global `parserOptions` is good idea, because technically webpack can generates and ECMA modules and any other module system

Oh that's nice! So the information coming from stats JSON should definitely be used here over a global option.

The first change could be to only change the `sourceType` and leave other stuff alone if it makes sense.

alexander-akait:
@eamodio Do you want to finish it, thank you?

eamodio:
Updated

valscion (commented):
Thanks! I think this direction looks safe and seems to be backwards-compatible.

Can you modify the PR title, description and README.md to describe what has now changed?

For example, something like this:

"Parse bundles as ES modules based on stats JSON information"

eamodio:
@valscion Updated the title & description, but not sure what you wanted added to the README as there isn't any new options or anything. But feel free to edit this PR as you see fit.

valscion:
Ah sorry, I meant the changelog! My bad.

eamodio:
Updated -- hopefully this is ready now

alexander-akait (approved):

valscion:
Can you avoid changing the formatting on all rows in the changelog? The diff there should be minimal.

Thank you for your patience with this PR. I'm mostly reviewing this on my phone so I can't easily modify the changes myself.

eamodio:
Ugh, sorry, damn auto-formatter. Fixed

valscion (approved):
Looks great, thank you! ☺️

PR #650, opened 2025-02-19
Drop support for Node versions less than 16.20.2

valscion:
The CI is failing as the list of node versions we have is so old that testing becomes difficult.

This PR drops support for EOL'd node versions to get CI passing again and make development of this library easier for outside contributors.

The older Node.js versions might still work — we are merely no longer testing against them.

There were seemingly some small regressions in some tests that had to be skipped or assertions changed to match the current reality. I've left PR review comments for every one of them for future reference. I think it's better to have a working CI than spend more time trying to tackle those test cases — especially as it seems the tests might've regressed already some time ago in history when CI was not working properly.


valscion:
OK not that many failing tests left. Some of the test failures themselves seem to be "real" test failures that had been masked by the CI not being in a working condition for some time...

valscion:
Huh. So tests on node.js v18.x are hanging for some mysterious reason. And I can't reproduce that locally...

valscion (commented):
Some potential regressions were highlighted as PR review comments.

valscion:
> This PR breaks local dev for me. The browser doesn't show the viewer but instead an error in the console:

Thank you for letting me know!

.babelrc
```diff
@@ -3,11 +3,7 @@
 {
   "presets": [
     ["@babel/preset-env", {
-      "targets": {"node": "10.13.0"}
+      "targets": {"node": "16.20.2"}
     }]
-  ],
-  "plugins": [
-    ["@babel/plugin-proposal-class-properties", {"loose": true}],
```
valscion, line 10: I suppose this `@babel/plugin-proposal-class-properties` was necessary after all and its removal caused the error you were seeing @dcsaszar. Maybe we should be using https://babeljs.io/docs/babel-plugin-proposal-decorators instead with some option that makes sense?

I'm not too familiar with the view code that uses these decorators. But we'd need to resolve that problem before I can release a new major version.

valscion: The mobx integration that is used in different places in the UI code seem to rely on decorators. I'm not too familiar with them but they probably need some looking into to make this library still work.

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/544a669a6a13d5e020a1be790171a1f00dee787c/client/store.js#L9-L14

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/544a669a6a13d5e020a1be790171a1f00dee787c/client/components/ModulesTreemap.jsx#L35-L36

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/544a669a6a13d5e020a1be790171a1f00dee787c/client/components/ModulesTreemap.jsx#L177

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/544a669a6a13d5e020a1be790171a1f00dee787c/client/components/ModulesTreemap.jsx#L181

https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/544a669a6a13d5e020a1be790171a1f00dee787c/client/components/ModulesTreemap.jsx#L185

blob/544a669a6a13d5e020a1be790171a1f00dee787c/client/components/ModulesTreemap.jsx#L198

blob/544a669a6a13d5e020a1be790171a1f00dee787c/client/components/ModulesTreemap.jsx#L202

PR #666, opened 2025-07-03
fix: allow react native bundle files

ilteoood:
Thanks to Re-Pack, now even React Native projects can use Rspack or Webpack as a bundler instead of Metro.
We would like to use webpack-bundle-analyzer to discover the pain point of our app.
Unfortunately, the React Native assets are stored in a `.bundle` size: this is forcing us to patch `webpack-bundle-analyzer` to be able to use it.

With this PR, I'm going to address it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 * **New Feature**
   * Add support for Brotli compression ([#663](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/663) by [@dcsaszar](https://github.com/dcsaszar))
+  * Add support for React Native ([666](https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/666) by [@ilteoood](https://github.com/ilteoood))
 
 ## 4.10.2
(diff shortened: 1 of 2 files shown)


ilteoood:
> Hmmyeah makes sense. Can you add a changelog entry as well?

Sure thing! Let me know if that looks good to you

ilteoood:
Thank you so much @valscion!
Do you know, more or less, when it will be released? I see that the last version is more than a year old now

valscion:
I would be able to release a new major version once some skipped tests have been debugged why they have stopped working or fix them so that the tests pass afain.

If you are able to help me debug why we now need to have `it.skip` on some tests, there would be no blockers on releasing a new version. I've described the skipped tests here: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/650#discussion_r1964157439

I'm right now on vacation from work so am sometimes online. But I can ship a new version regardless once those tests have been fixed.

valscion:
@ilteoood try v5.0.1, it includes this change ☺️ hopefully it would work for you!

ilteoood:
Thanks for the ping @valscion, it works quite well 👌

PR #677, opened 2025-11-07
Upgrade packages

valscion:


valscion:
Doesn't look too easy. Don't have time to look into simplifying the gulp workflow now.

PR #683, opened 2025-12-09
Dark mode toggle implementation

theEquinoxDev:
**Summary**

This PR adds an optional dark/light mode toggle for the webpack-bundle-analyzer UI, addressing Issue #642.

**Implementation**

-Added a theme toggle button in the interface
-Implemented CSS custom properties for dark/light theme colors
-Stored theme preference using localStorage
-Updated UI components to support both modes

**What kind of change does this PR introduce?**

Type: Feature addition (new functionality)
Scope: User interface enhancement with theme switching capability

**Tests**
No tests added in this PR — functionality affects UI/theme only.

**Breaking changes**
None. Dark mode is optional, defaults to the existing theme.

This PR addresses the exact issue described in #642 by providing a professional dark mode implementation that enhances user experience without disrupting existing functionality. The implementation follows the project's existing patterns and maintains visual consistency across all interface elements.

**Screenshots**

-Dark Theme
image
-Light Theme
image

**Notes**
Happy to update styling or toggle placement based on feedback. 
Thanks for the opportunity to work on this feature!

### DIFF
--- a/client/store.js
+++ b/client/store.js
@@ -14,15 +14,17 @@ export class Store {
   @observable showConcatenatedModulesContent = (localStorage.getItem('showConcatenatedModulesContent') === true);
   @observable darkMode = (() => {
     const systemPrefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-  
+
     try {
       const saved = localStorage.getItem('darkMode');
       if (saved !== null) return saved === 'true';
-    } catch (e) {}
-  
+    } catch (e) {
+      // Some browsers might not have localStorage available and we can fail silently
+    }
+
     return systemPrefersDark;
   })();
-  
+
 
   setModules(modules) {
     walkModules(modules, module => {
@@ -196,6 +198,7 @@ export class Store {
     try {
       localStorage.setItem('darkMode', this.darkMode);
     } catch (e) {
+      // Some browsers might not have localStorage available and we can fail silently
     }
     this.updateTheme();
   }
--- a/client/components/ThemeToggle.jsx
+++ b/client/components/ThemeToggle.jsx
@@ -12,8 +12,7 @@ export default class ThemeToggle extends Component {
     const {darkMode} = store;
 
     return (
-      <Button
-        type="button"
+      <Button type="button"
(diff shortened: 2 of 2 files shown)


client/store.js
```diff
@@ -12,6 +12,7 @@ export class Store {
   @observable defaultSize;
   @observable selectedSize;
   @observable showConcatenatedModulesContent = (localStorage.getItem('showConcatenatedModulesContent') === true);
+  @observable darkMode = (localStorage.getItem('darkMode') === true);
```
valscion: The browser can disable access to `localStorage` in some environments so we should protect against `getItem` or `setItem` throwing an error. We can just fail silently and assume that dark mode is unset and thus use the system value as the default.

valscion (commented):
This looks like a good start, thanks!

Could we default the dark mode value to what the system has decided and then allow the user to configure something else? That is, the initial value would be based on what `prefers-color-scheme` media query says: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-color-scheme

And once done, let's also add a changelog entry.

Bonus points if you can figure out some way of having a test for this — I don't think we have any UI-level tests so it's also fine to leave testing out of this PR.

theEquinoxDev:
Thanks for the review! I'll update the implementation to:
-use prefers-color-scheme as the default
-only apply localStorage if available
-add a changelog entry

will push shortly..

theEquinoxDev:
Thanks for the feedback. I have updated the implementation to:
-Use the system prefers-color-scheme value as the initial state
-Add a safe fallback for localStorage access
-Insert a changelog entry under the UNRELEASED section

Please let me know if anything else should be adjusted. Thanks for the opportunity.

valscion:
Released in v5.1.0 ☺️

theEquinoxDev:
Thanks for the approval and feedback! Really glad the feature has been useful. I'll definitely keep an eye on the dark mode CSS and accessibility improvements and would be happy to iterate further if needed.
Appreciate the opportunity to contribute 😊

PR #685, opened 2025-12-15
fix: failed to parse bundle with IIFE

hai-x:
**Summary**

**What kind of change does this PR introduce?**

Fixes https://github.com/webpack/webpack-bundle-analyzer/issues/684, https://github.com/webpack/webpack-bundle-analyzer/issues/409.

**Did you add tests for your changes?**

It seems we currently don’t support parsing top-level chunk modules and i’m also not sure whether parsing is the right approach for get rendered chunk modules.

e.g.
```
const __webpack_modules__ = [,,,...]
```

So i add one simple IIFE case here without chunk modules.

Here is test result before fixing: 
image

**Does this PR introduce a breaking change?**

No

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _Note: Gaps between patch versions are faulty, broken or test releases._
 
 * **Bug Fix**
   * Fix tooltip styling in dark mode when using CSS Modules ([#688](https://github.com/webpack/webpack-bundle-analyzer/pull/688) by [@theEquinoxDev](https://github.com/theEquinoxDev))
-
+  * Avoid parse failures for bundles with IIFE ([#685](https://github.com/webpack/webpack-bundle-analyzer/pull/685) by [@hai-x](https://github.com/hai-x))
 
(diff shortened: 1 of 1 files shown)


alexander-akait:
@evenstensberg We need to add test cases

alexander-akait (approved):

alexander-akait:
@valscion In this case there are no modules, anyway in our TODOs to add all test cases for ES modules with and without IIFE, it is just an edge case

valscion:
Yeah that's true. I'll add a changelog entry myself when I have the time or then @hai-x can add one — I'll merge once the changelog is there and release when I have a good slot :)

hai-x:
@valscion I'm so sorry, I missed the message and forgot to reply. Thank you for your help. ❤️

valscion:
Sure, not a problem @hai-x! ☺️

valscion:
Released in v5.1.1. Thanks for the contribution!

Issue #490, opened 2022-01-23
TypeError: Cannot read property 'forEach' of undefined


sahariko:
#### Issue description

When running the CLI on a stats file (not all stats files, just some), I'm getting the following error, and the following error only:

```
Could't analyze webpack bundle:
TypeError: Cannot read property 'forEach' of undefined
```

It's not a lot to go on, there aren't any helpful hints or file names in which it crashes.

#### Technical info

<pre>
System:
    OS: macOS 11.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 3.08 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.22.0 - ~/.nvm/versions/node/v12.22.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v12.22.0/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v12.22.0/bin/npm
  npmPackages:
    terser-webpack-plugin: 5.1.3 => 5.1.3 
    webpack: 5.64.1 => 5.64.1 
    webpack-cli: 4.7.2 => 4.7.2 
    webpack-dev-server: 4.7.3 => 4.7.3 
    webpack-manifest-plugin: 3.1.1 => 3.1.1 
</pre>

#### Debug info

Using the CLI - `npx webpack-bundle-analyzer@latest stats.json`, with the following plugins:
- [`@sentry/webpack-plugin`](https://github.com/getsentry/sentry-webpack-plugin)
- [`terser-webpack-plugin`](https://webpack.js.org/plugins/terser-webpack-plugin/)
- [`webpack-manifest-plugin`](https://github.com/shellscape/webpack-manifest-plugin)
- [`mini-css-extract-plugin`](https://github.com/webpack-contrib/mini-css-extract-plugin)
- [`terser-webpack-plugin`](https://webpack.js.org/plugins/terser-webpack-plugin/)
- [`webpack.ContextReplacementPlugin`](https://webpack.js.org/plugins/context-replacement-plugin/)
- [`webpack.DefinePlugin`](https://webpack.js.org/plugins/define-plugin/)

valscion:
> on a stats file (not all stats files, just some),

Are you able to post a stats file that causes this issue?

[...]

PR #717, opened 2026-06-04
fix: handle child stats without root assets

puneetdixit200:
## What

Initialize the selected child stats object's `assets` array before appending assets from later child compilations.

## Why

When root stats contain only `children`, and the first child has no `assets` while a later child does, `getViewerData()` currently tries to push into `undefined` and crashes before producing chart data. This is one of the missing-stats-shape cases discussed in #490.

## How

Added a focused regression test for that stats shape and guarded the selected child stats with an empty `assets` array before merging later child assets.

## Validation

- `NODE_OPTIONS=--openssl-legacy-provider npx jest test/analyzer-get-viewer-data.js --runInBand` failed before the fix with `Cannot read properties of undefined (reading 'push')`
- `NODE_OPTIONS=--openssl-legacy-provider npx jest test/analyzer-get-viewer-data.js --runInBand`
- `npm run build`
- `NODE_OPTIONS=--openssl-legacy-provider npx jest test/analyzer-get-viewer-data.js test/analyzer.js --runInBand --testTimeout=60000`
- `npm run lint`
- `git diff --check`

Note: I also tried the full Jest suite with `--testTimeout=60000`; the analyzer tests passed, but the run failed in `test/plugin.js` after 128 passed / 4 skipped due to a local Puppeteer `afterEach` browser cleanup timeout.

### DIFF
--- a/src/analyzer.js
+++ b/src/analyzer.js
@@ -230,6 +230,7 @@ function getViewerData(bundleStats, bundleDir, opts) {
   ) {
     const { children } = bundleStats;
     [bundleStats] = bundleStats.children;
+    bundleStats.assets ||= [];
     // Sometimes if there are additional child chunks produced add them as child assets,
     // leave the 1st one as that is considered the 'root' asset.
     for (let i = 1; i < children.length; i++) {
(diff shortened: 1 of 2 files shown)


alexander-akait:
Please accept CLA and fix copilot review

PR #718, opened 2026-06-12
fix: request analyzer stats fields

SyMind:
## Summary

Rspack 2.0 changed the default `stats.toJson()` output so fields like `modules`, `assets`, `chunks`, `chunkGroups`, and `entryPoints` are no longer included unless requested explicitly. See the [Rspack v2 migration guide](https://rspack.rs/guide/migration/rspack_1.x#changed-default-parameters-of-statstojson).

That can leave webpack-bundle-analyzer with incomplete stats when it runs against Rspack 2.0. This PR makes analyzer modes request the stats fields they need directly instead of relying on compiler defaults.

## Changes

- Add internal stats options for bundle analysis.
- Use them in `server`, `static`, and `json` analyzer modes.
- Keep user-provided `statsOptions` scoped to generated stats files.
- Add regression tests and a patch changeset.

## Notes

`statsOptions` remains user-controlled because users may intentionally filter fields such as assets or modules. Analyzer report generation needs a stable minimum stats shape, so it uses dedicated internal options.

## Validation

- `npm test`
- `npm run lint`
- `npm run build:analyzer`
- `npm run lint:types`
- Targeted Jest, ESLint, and Prettier checks for the touched files.

### DIFF
--- a/.changeset/fair-forks-tap.md
+++ b/.changeset/fair-forks-tap.md
@@ -2,4 +2,4 @@
 "webpack-bundle-analyzer": patch
 ---
 
-Ensure analyzer modes request the stats fields needed for bundle analysis.
+Explicitly define `stats.toJson()` options for bundle analysis to support Rspack 2.0 with minimal code changes needed.
--- a/.changeset/fair-forks-tap.md
+++ b/.changeset/fair-forks-tap.md
@@ -1,0 +1,5 @@
+---
+"webpack-bundle-analyzer": patch
+---
+
+Ensure analyzer modes request the stats fields needed for bundle analysis.
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -30,6 +30,32 @@ const viewer = require("./viewer");
 /** @typedef {string | (() => string)} ReportTitle */
 /** @typedef {(options: { listenHost: string, listenPort: number, boundAddress: string | AddressInfo | null }) => string} AnalyzerUrl */
 
+/** @type {StatsOptions} */
+const analyzerStatsOptions = {
+  all: false,
+  assets: true,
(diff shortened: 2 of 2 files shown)


valscion (commented):
What use case needs this? Never heard of this kind of need in the past

SyMind:
@valscion This is mainly for Rspack 2.0 compatibility.

Rspack 2.0 changed the default output of `stats.toJson()`: fields such as `modules`, `assets`, `chunks`, `chunkGroups`, and `entryPoints` are no longer included by default unless requested explicitly:
https://rspack.rs/guide/migration/rspack_1.x#changed-default-parameters-of-statstojson

webpack-bundle-analyzer currently calls `stats.toJson()` without options in analyzer modes, so with Rspack 2.0 it can receive incomplete stats and fail to build the expected bundle analysis data.

The goal of this PR is to make the analyzer request the minimum stats shape it depends on, instead of relying on compiler-specific defaults.

valscion (approved):
Hmm OK, thanks for explaining! I do understand this better now.

Now I wonder if there's any way to add reasonable test coverage for this... we don't really support Rspack in the sense of having any guarantees of it working here nor do we want to have CI test against it as it brings additional maintenance burden to verify it doesn't break here.

But this is such a small change and doesn't seem brittle so might as well have it. I'll modify the changeset a bit to describe this stance on Rspack.

Verthon:
Hello @valscion is it possible to try this fix on some alpha/custom version of the webpack-bundle-analyzer ? Thanks

valscion:
https://github.com/webpack/webpack-bundle-analyzer/issues/720 will likely need help with first :)

valscion:
Turns out CI was happy now so I released v5.3.1 with this change in :)

PR #725, opened 2026-07-26
fix: preserve compiler paths for shared plugin instances

### LINKED ISSUE #506: Incorrect bundle paths when bundles are in multiple directories.
#### Issue description

Where bundles are output to more than one directory, bundle paths are incorrect. 

I'm using `webpack-bundle-analyser` with Gatsby via [gatsby-plugin-webpack-bundle-analyser-v2](https://github.com/JimmyBeldone/gatsby-plugin-webpack-bundle-analyser-v2) which is a very thin plugin. Gatsby adds bundles to `/public/`, however another plugin I use [gatsby-plugin-netlify-cms](https://www.gatsbyjs.com/plugins/gatsby-plugin-netlify-cms/) adds bundles to `/public/admin/`. When `BundleAnalyzerPlugin` [derives the `bundleDir`](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/7d6039e24e84c6e156a8dea36528a12a554bb714/src/BundleAnalyzerPlugin.js#L136) it resolves (using `this.compiler.outputPath` to `/path/to/public/admin/` which it then uses as a prefix for all bundle paths. This means only the bundles that are generated to `/admin/` are correct. The rest are not and result in warnings:

> Error parsing bundle asset "/Users/me/Sites/www/public/app.js": no such file

It seems the current architecture can't handle situations where bundles are in two different locations. 

#### Technical info

<pre>
  System:
    OS: macOS 12.3.1
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 133.81 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 14.18.2 - ~/.nvm/versions/node/v14.18.2/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.2/bin/npm
    Watchman: 2021.11.15.00 - /usr/local/bin/watchman
  npmPackages:
    gatsby-plugin-webpack-bundle-analyser-v2: ^1.1.26 => 1.1.26 </pre>
</pre>

utkarshalpha:
## Summary

Fixes #506.

A `BundleAnalyzerPlugin` instance can be applied to multiple compilers, but the plugin previously stored only the most recently applied compiler. As a result, every done-hook callback resolved bundle assets and report files against that compiler's output directory.

This change:

- passes the compiler captured by each done-hook closure into report and stats generation;
- resolves bundle directories and output filenames against that compiler;
- uses the current compiler's bundle directory when refreshing server-mode chart data; and
- adds a multi-compiler regression test with one shared plugin instance for both Webpack 4 and Webpack 5.

## Testing

- `npm exec -- jest test/plugin.js --runInBand --testNamePattern=should.use.each.compiler.output.path` (2 passed)
- `npm exec -- jest test/viewer.js --runInBand` (5 passed)
- `npm exec -- eslint src/BundleAnalyzerPlugin.js src/viewer.js test/plugin.js`
- `npm exec -- tsc --pretty --noEmit`
- `npm exec -- prettier --check src/BundleAnalyzerPlugin.js src/viewer.js test/plugin.js .changeset/calm-compilers-report.md`
- `npm run build`

### DIFF
--- a/src/BundleAnalyzerPlugin.js
+++ b/src/BundleAnalyzerPlugin.js
@@ -127,7 +127,10 @@ class BundleAnalyzerPlugin {
 
       if (this.opts.generateStatsFile) {
         actions.push(() =>
-          this.generateStatsFile(stats.toJson(this.opts.statsOptions)),
+          this.generateStatsFile(
+            stats.toJson(this.opts.statsOptions),
+            compiler,
+          ),
         );
       }
 
@@ -138,15 +141,21 @@ class BundleAnalyzerPlugin {
 
       if (this.opts.analyzerMode === "server") {
         actions.push(() =>
-          this.startAnalyzerServer(stats.toJson(analyzerStatsOptions)),
+          this.startAnalyzerServer(
+            stats.toJson(analyzerStatsOptions),
+            compiler,
+          ),
         );
       } else if (this.opts.analyzerMode === "static") {
         actions.push(() =>
-          this.generateStaticReport(stats.toJson(analyzerStatsOptions)),
+          this.generateStaticReport(
+            stats.toJson(analyzerStatsOptions),
+            compiler,
+          ),
         );
       } else if (this.opts.analyzerMode === "json") {
         actions.push(() =>
-          this.generateJSONReport(stats.toJson(analyzerStatsOptions)),
+          this.generateJSONReport(stats.toJson(analyzerStatsOptions), compiler),
         );
       }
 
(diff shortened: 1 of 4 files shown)


src/BundleAnalyzerPlugin.js
```diff
@@ -175,12 +184,15 @@ class BundleAnalyzerPlugin {
 
   /**
    * @param {StatsCompilation} stats stats
+   * @param {Compiler=} compiler compiler
    * @returns {Promise<void>}
    */
-  async generateStatsFile(stats) {
+  async generateStatsFile(
+    stats,
+    compiler = /** @type {Compiler} */ (this.compiler),
```
valscion: Is there a reason to allow leaving out the `compiler` argument now? This looks like AI using protective coding pattern over understanding if this default is what we want or not.

utkarshalpha: Good catch — there was no real need for the default. The hook already threads the actual compiler through every internal call site, so I removed the `this.compiler` fallback and made the parameter required in 8fabe1a.

valscion (approved):
Looks good to me!

Also, noticed that `analyzerMode === "server"` is not really covered even in base branch so those changed lines also can't be verified by CI... would be lovely to see a change that would add missing coverage for that.

PR #734, opened 2026-08-23
fix: scope parsed module sources by asset

### LINKED ISSUE #732: Parsed sources collide across assets, and getViewerData mutates input stats
## Problem

Current `main` (`23adc7a`) has two independent `getViewerData()` correctness issues.

### Parsed-source lookup is not asset-scoped

Verified with Node 24.18.0 and Webpack 5.105.2:

- root asset `app.js` and child-compilation asset `worker.js` both used module ID `0`;
- their parsed sources were 60 and 182 bytes;
- the `app.js` module received the worker source, reporting 182 parsed bytes despite `app.js` being 116 bytes.

The public `test/stats/with-worker-loader` fixture also produced two incorrect same-ID source assignments.

`getViewerData()` currently combines parsed module maps from every asset, keyed only by module ID. IDs reused by separate compilations can therefore overwrite each other.

### Analysis mutates the supplied stats

The same reproduction showed that `getViewerData()`:

- replaced and grew the root assets array from one item to two;
- removed an asset query string in place;
- appended the child asset and set `isChild`;
- added `parsedSrc` to the root module.

Calling it again grew the assets array from two items to three.

## Expected behavior

- Resolve parsed sources from the current asset’s module map.
- Produce chart data without modifying the supplied stats object.
- Return stable results when called repeatedly with the same input.

## Proposed follow-up PRs

1. **Asset-scoped parsed-source lookup**
   - Retain parsed module maps per asset.
   - Add regressions for root and child assets that reuse a module ID.
   - Keep input-mutation changes out of scope.

2. **Non-mutating analysis**
   - Stop modifying caller-owned asset arrays, asset records, child assets, and modules.
   - Add immutability and repeated-call regressions.
   - Keep parsed-source lookup and unrelated analyzer changes out of scope.

[PR #723](https://github.com/webpack/webpack-bundle-analyzer/pull/723) changes module selection immediately above the parsed-source loop. Its optimization is semantically separate, but these follow-ups will likely need a textual rebase.

Menny1337:
Fixes #732.

## Problem

Parsed module sources were attributed globally in two ways.

**1. Module IDs were treated as globally unique.** `getViewerData()` merged every asset's parsed
module map into one object:

```js
Object.assign(parsedModules, bundleInfo.modules);
```

Webpack module IDs are unique only within a compilation. Separate compilations can reuse the same
IDs. The last parsed asset then overwrote sources from earlier assets.

**2. Modules shared by several assets were changed in place.** Each asset received the same stats
object for a shared module. The analyzer wrote `parsedSrc` to that object. The module tree read the
value after all assets were processed, so the last asset set the value for every asset.

Both cases made modules report parsed and compressed sizes from another asset.

## Fix

- Store parsed sources by asset name. Each asset reads only its own module map.
- Use `Object.hasOwn` for asset and module lookups.
- Copy the module objects for each asset before writing `parsedSrc`.

The asset and module indexes added in #723 remain unchanged. Lookup complexity remains linear.

## Regression tests

The tests cover both failure modes.

- **Duplicate IDs across compilations:** The existing
  `test/stats/with-worker-loader-dynamic-import` fixture has a root bundle and a worker bundle.
  Both compilations use module ID `0` for different modules.
- **One module in two assets:** The new `test/stats/with-module-in-multiple-assets` fixture comes
  from a Webpack 5.105.2 production build. Two entry files import different exports from
  `messages.js`. Webpack reports the source size as `87` for both assets, but tree shaking creates
  factories of 59 and 37 characters.

The new fixture includes its source files, Webpack configuration, generated bundles, reduced
stats, and a README.

A separate test covers the fallback used when the analyzer cannot parse any bundle. The local
coverage report covers all changed statements and branches in `src/analyzer.js`.

`test/analyzerUtils.js` now checks the returned chart data instead of depending on mutation of the
input stats.

## Notes

- Patch changeset included.
- This PR does not change the remaining mutation of the `bundleStats` asset array, asset names, or
  `isChild` flags.

## AI assistance

This change was developed with meaningful AI assistance for investigation, patch drafting, and
test authoring. I reviewed and verified all code, tests, and validation output before submitting.


valscion (commented):
Thanks! Can you explain what kind of source code and webpack configuration produces these test bundles and the stats.json? I'm not sure I follow the test fixture logic here. The stats.json seems misleasing already from the webpack side? The shared module seems like it should have same size in both bundles if stats.jsln stat size is to be believed but the bundle contents differ for the shared module.

valscion (approved):
Thanks, this fell through the cracks. All looks good! The webpack generated real bundle is the true gold nugget in here and I can now understand how the error has happened.

Nice fix! Thanks for the contributions you've done here, I appreciate it!

